PR #11, opened 2023-02-23
Extract parser to Coq

### LINKED ISSUE #9: Extract parser to Coq
c.f. #2 for command and examples.

jinh0:
Closes #9  .

Comments:
- I'm not allowed to use `Type` as a constructor in our `Cst` type since `Type` seems to be a reserved keyword in Coq.

Todo:
- [x] For some reason, `Parser.v` is giving me an error saying that the parser is not complete. Not sure why.
```
File "./Parser.v", line 1342, characters 6-64:
Error:
The term "eq_refl" has type "true = true" while it is expected to have type
 "MenhirLibParser.complete_validator tt = true".
```
- [x] Extract OCaml code from `Parser.vy`


HuStmpHrrr:
The parser is not complete because the grammar has conflict: http://gallium.inria.fr/~fpottier/menhir/manual.html#sec86

The conflict is caused by application. Imagine tokens `a b c`, the current parsing rule allows `(a b) c` and `a (b c)`. On paper, this conflict is resolved by saying application is left associative, but unfortunately, the manual says the Coq backend doesn't support it. If my memory is correct, any conflict of this kind can be resolved by changing the parsing rules. Have you taken any class on compilers or parsing?

HuStmpHrrr:
A random search online shows that people do resolve this conflict by associativity:

https://github.com/chrisnevers/lambda-calculus/blob/c37987bac024cc01b894f8dde924a2f152b65c5f/src/parser.mly#L28

I don't exactly remember how to resolve this conflict by changing the parsing rule.  Try to look into LR(1) parsing. If you can't figure it out easily, I will take a look after the paper deadline.

HuStmpHrrr:
this might work: https://github.com/utgwkk/lambda-chama/blob/master/parser.mly

jinh0:
Thanks for the help! The last link you sent me worked like a charm. No warnings or errors, Parser.vy compiles with no issue now.

lib/coq/Parser.vy
```diff
@@ -0,0 +1,76 @@
...
+
+(* M N *)
+app_obj:
+  (* simpl_obj prevents conflict by associativity *)
+  | app_obj simpl_obj { Cst.App $1 $2 }
+  | simpl_obj { $1 }
+
+(* Either a primitive object like "Type 5" / "Zero"
+  or parentheses around a complex object *)
+simpl_obj:
```
HuStmpHrrr, line 69: all cases other than VAR and parentheses can be placed in the obj rule, as none of the rest can be used as a function.

lib/coq/Parser.v
```diff
@@ -0,0 +1,1323 @@
+
```
HuStmpHrrr, line 1: since this file is generated, maybe add it to .gitignore?

HuStmpHrrr:
> Thanks for the help! The last link you sent me worked like a charm. No warnings or errors, Parser.vy compiles with no issue now.
> 
> If I understand my code correctly, you now need to explicitly give parentheses for applications like `M (N (O P))`.

that's great.

lib/coq/Extract.v
```diff
@@ -0,0 +1,3 @@
+(* Requires menhirlib: opam install coq-menhirlib *)
```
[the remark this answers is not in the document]
jinh0: Yeah, I had to install it separately. I'll update the README.

jinh0:
Hey @HuStmpHrrr I'm a little stuck on the extraction process:

- I keep running into this issue with the extracted parser expecting a `unit` argument for tokens like `LPAREN` and `RPAREN`.
- I don't know how to run the parser.

I wrote the extraction code as follows:

```coq
Require Extraction.
From McLtt Require Import Parser.

Import MenhirLibParser.Inter. (* I don't know why this is here *)

Extraction Language OCaml.

Extract Inductive nat => "int"
  [ "0" "(fun x -> x + 1)" ]
  "(fun zero succ n -> if n=0 then zero () else succ (n-1))".

Require Import ExtrOcamlString. (* Coq string to OCaml string, more like char list *)

Extract Inductive unit => "unit" [ "()" ].

Extraction "lib/parser.ml" Parser.prog.

```

I realized I need to convert the Coq `nat` type to `int`, and Coq `string` to OCaml's `char list` (see https://stackoverflow.com/a/25623226 ). No problem.

But, the Coq parser expects tokens like `LPAREN` to have an argument of type `unit0`, defined as just `type unit0 = Tt`.
So I used the line `Extract Inductive unit => "unit" [ "()" ].` in the extraction code so that I can convert `unit0` to OCaml's `unit`. But this doesn't solve the issue of the `unit` expected. So I added `unit` for every token affected in the lexer.
```
  | '(' { LPAREN () }
  | ')' { RPAREN () }
  | '.' { DOT () }
```

Okay, this seems weird.

But then I realized, I actually don't know how to run the parser.

The parse function takes these arguments:
```
# Parser.prog;;
- : int ->
    Parser.MenhirLibParser.Inter.buffer ->
    Parser.Cst.obj Parser.MenhirLibParser.Inter.parse_result
```

How do I generate this `Parser.MenhirLibParser.Inter.buffer` type?

HuStmpHrrr:
Good that you asked. Try
```coq
Require Coq.extraction.Extraction
```

the source code is here: https://github.com/coq/coq/tree/master/theories/extraction

I did not analyze the exact reason for your problem, but it most likely has been encountered before. This module is the standard module to require when you want to do an extraction.

HuStmpHrrr:
actually also try to include 

```
Require Import ExtrOcamlBasic.
Require Import ExtrOcamlNatInt.
Require Import ExtrOcamlString.
```

and possibly more if they still not fix it.

jinh0:
Okay, I added those lines of code, but that didn't change anything. The only issue I have left is that tokens that take no arguments still take `unit` as an argument, which I guess is okay.

Anyway, after doing some research on how these lazy buffers `Parser.MenhirLibParser.Inter.buffer` work, (see https://github.com/AbsInt/CompCert/blob/master/cparser/Lexer.mll#L601 , https://gitlab.inria.fr/fpottier/menhir/-/blob/master/demos/calc-incremental/calc.ml#L15 , https://gitlab.inria.fr/fpottier/menhir/-/blob/master/demos/coq-minicalc/Lexer.v#L60),
tada!
```ocaml
# open Mcltt;;
# open Parser.Cst;;
# Main.parse "succ zero";;
- : obj option = Some (Succ Zero)
# Main.parse "fun (x:Type 1) . x";;
- : obj option = Some (Fun (['x'], TType 1, Var ['x']))
# Main.parse "x";;
- : obj option = Some (Var ['x'])
# Main.parse "a ((b c) d)";;
- : obj option =
Some (App (Var ['a'], App (App (Var ['b'], Var ['c']), Var ['d'])))
```

Main thing was to make a function that generates a lazy buffer from the lexbuf:
```ocaml
let rec loop x = lazy (Buf_cons (Lexer.read x, loop x))

let parse s =
  let lexbuf = Lexing.from_string s in
  let buffer = loop lexbuf in
  match Parser.prog 50 buffer with
  | Parsed_pr (e, _) -> Some e
  | Fail_pr_full (_, _) -> None
  | _ -> None
```

Todo:
- [ ] I do have to fix the tests in `test.ml` so that they actually pass.
- [ ] I actually messed up the order of arguments in multiple-args functions, so I'll fix that.

HuStmpHrrr:
that's great! I am glad that you are taking a look into compcert, including other references, yourself.

jinh0:
Okay, I edited the tests and they now all work with the new Coq parser! The tests actually helped me catch two bugs.

Since `parser.ml` and `parser.mli` are generated by the extraction and are disgustingly long, I've added them to the .gitignore and added a `Makefile` with a command `gen_parse` to build and extract the Coq parser.

I think everything is done for this pull request, except two minor caveats:
- The Cst takes in char list rather than strings.
- Tokens like `LPAREN` take in the `unit` argument => `LPAREN ()`

HuStmpHrrr:
I don't think char list is a big deal. If you really want to, try to require https://github.com/coq/coq/blob/master/theories/extraction/ExtrOcamlNativeString.v and see if everything compiles.

otherwise, well done!

lib/test.ml
```diff
@@ -1,28 +1,27 @@
...
-let%test "lambda" = parse "fun (x : Type 5).y" = Fun ("x", Type 5, Var "y")
-let%test "lambda multiple args" = parse "fun (x : Nat) (y : Nat) . x" = Fun ("x", Nat, Fun ("y", Nat, Var "x"))
-let%test "lambda multiple args 2" = parse "fun (x : Nat) (y : Nat) (z : Nat) . z" = Fun ("x", Nat, Fun ("y", Nat, Fun ("z", Nat, Var "z")))
-let%test "application" = parse "(fun (x : Nat).x) Nat" = App (Fun ("x", Nat, Var "x"), Nat)
-let%test "nested 1" = parse "(Type 5) zero" = App (Type 5, Zero)
-let%test "nested 2" = parse "succ (succ (succ (succ zero)))" = Succ (Succ (Succ (Succ Zero)))
-let%test "pi" = parse "pi (x:Nat).x" = Pi ("x", Nat, Var "x")
-let%test "pi multiple args" = parse "pi (x : Nat) (y : Nat) (z : Nat) . z" = Pi
-("x", Nat, Pi ("y", Nat, Pi ("z", Nat, Var "z")))
+let x, y, z = ['x'], ['y'], ['z']
```
HuStmpHrrr, line 6: is there any actual change to the tests?

jinh0: No, effectively the tests are the same. Sorry, I should have clarified in the commit message, all I fixed was handling the option type output of `parse` and replace strings with char lists.

jinh0:
> try to require https://github.com/coq/coq/blob/master/theories/extraction/ExtrOcamlNativeString.v and see if everything compiles.

Huh, don't know how I missed `ExtrOCamlNativeString`. I added it and it works as expected! Thanks!

HuStmpHrrr:
good! everything looks OK to me. is there anything pending in this PR?

jinh0:
Nope, everything is done and working!

PR #12, opened 2023-03-14
[WIP] Well-Scopedness Lemma

IanGrenville:
The proof is still currently incomplete. 
I'm specifically stuck on the function case at this point, I've managed to get the others to go through.
There's also a lemma which seems both necessary and simple, but I haven't quite managed to prove yet.


lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+  | cNat => StrSet.empty
+  | cZero => StrSet.empty
+  | cSucc c => cst_variables c
+  | cVar s => StrSet.singleton s
+  | cFun s c => StrSet.remove s (cst_variables c)
+  |cApp c1 c2 => StrSet.union (cst_variables c1) (cst_variables c2)
  end
 .
 
+Fixpoint max_index (ast : ast_term) : option nat :=
```
HuStmpHrrr, line 84: the following 3 definitions are not optimal. consider the `close_at` predicate I described in the meeting

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+Fixpoint closed_at (ast : ast_term) (n : nat) : Prop :=
+match ast with
+  | aVar m => m < n
+  | aFun a => closed_at a n
+  | aApp a1 a2 => and (closed_at a1 n) (closed_at a2 n)
+  | _ => True
+end.
+
+
+Lemma elab_concat (cst : cst_term) (ctx : list string) (s : string) (x : elab_success cst ctx) : elab_success cst (s::ctx).
```
HuStmpHrrr, line 123: not needed.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+  (*Another situation where I'm stuck and not sure how I can use the induction
+    hypothesis that's available. *)
+
+Admitted. 
+(*Lemma for the older version of the well_scoped proof, not needed for the inductive predicate version*)
+Lemma lookup_known (s : string) (ctx : list string) (p : List.In s ctx) : exists n : nat, (lookup s ctx = Some n).
+Proof.
+  
+  
+  induction ctx as [| c ctx' IHctx].
```
HuStmpHrrr, line 138: I am almost sure you can do an induction on `p` and get a simpler proof. In general, we use `H` something for hypothesis that lives in `Prop`.  the "something" usually is something more meaningful.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+  exists(x0 + 1).
+  rewrite H3.
+  reflexivity.
+Qed.
+
+(*Alternate phrasing of the lemma using an inductive predicate for
+  elaboration suceeding, instead of the "exists a, elaborate ... = Some a"
+  I think it should be equivalent, but for this one most steps go through easier.
+*)
+Lemma alt_well_scoped (cst : cst_term) (ctx : list string)
```
HuStmpHrrr, line 186: this one is less good than the one below.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+  
+
+  
+  (*Admitted just so the lower proof attempt isn't blocked*)
+Admitted.
+
+
+(* If all free variables are in a context, then elaboration suceeds with that context*)
+Lemma well_scoped (cst : cst_term) (ctx : list string) 
+(x : StrSet.Subset (cst_variables cst) (StrSProp.of_list ctx)) :
```
HuStmpHrrr, line 271: utilize some notation. I am almost sure there must be something around in the library already.

HuStmpHrrr: c.f. https://github.com/coq/coq/blob/master/theories/MSets/MSetInterface.v#L145

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+(*Alternate phrasing of the lemma using an inductive predicate for
+  elaboration suceeding, instead of the "exists a, elaborate ... = Some a"
+  I think it should be equivalent, but for this one most steps go through easier.
+*)
+Lemma alt_well_scoped (cst : cst_term) (ctx : list string) 
+(x : StrSet.Subset (cst_variables cst) (StrSProp.of_list ctx)) :
+ elab_success cst ctx.
+Proof.
+  induction cst.
+  all : simpl ; trivial.
```
HuStmpHrrr, line 191: if this line does do anything, you can combine it with the previous one using a `;`.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+  simpl in x.
```
HuStmpHrrr, line 192: at this point, there should be multiple goals. use `-` to enter a subgoal. use `+` and `*` for more nested goals.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+  assert (StrSet.In s (StrSProp.of_list ctx)).
+  apply x.
+  apply (StrSet.singleton_spec s s).
+  reflexivity.
+  destruct ( (StrSProp.of_list_1 ctx s)).
+  destruct (InA_alt eq s ctx).
```
HuStmpHrrr, line 198: I am not sure what this is. you most likely poke into a definition which you don't need to look into. consider searching a lemma.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
 
 
 (**De-monadify with pattern matching for now**)
-Fixpoint lookup (s : string) (ctx : list string) (index : nat) : (option nat) :=
+Fixpoint lookup (s : string) (ctx : list string) : (option nat) :=
  match ctx with
   | nil => None
-  | c::cs => if eqb c s then (Some index) else lookup s cs (index+1) 
+  | c::cs => if String.eqb c s then (Some 0) else
```
HuStmpHrrr, line 38: instead of `eqb`, use `string_dec`. you can still use `if`. The advantage of the latter is that it provides more convenient information when you prove things.  I think we can control it and make it to extract to boolean eventually.

lib/elaborator.v
```diff
@@ -22,42 +32,295 @@ Inductive ast_term : Type :=
...
+     Still gets stuck on essentially the same step as without looking at the cases
+  destruct cst.
+  all : simpl ; trivial.
+  destruct (String.string_dec s s0).
+  left.
+  exact e.
+  right.
+  
+  simpl in x.
+  unfold StrSet.Subset in x.
```
HuStmpHrrr, line 243: most likely this line is not needed. whatever you are trying to do should be covered by a lemma.

HuStmpHrrr: look into https://coq.github.io/doc/v8.11/stdlib/Coq.MSets.MSetFacts.html

IanGrenville:
@HuStmpHrrr I think I've made all the changes we've discussed so far, and I've been able to get the proof to go all the way through in its current state. 

What I will still need to do is update it to use the slightly different AST and CST definitions we have in Syntax.v, which will include adding a few more cases to the proof.

lib/elaborator.v
IanGrenville: When I tried doing induction on the hypothesis I got an error `Not an inductive product.`

HuStmpHrrr:
Ok that's great. I will read in more details.

lib/elaborator.v
HuStmpHrrr: hmm I will check how `In`  is defined.

theories/Elaborator.v
```diff
@@ -0,0 +1,340 @@
+Require Export String.
```
HuStmpHrrr, line 1: add this file to `_CoqProject` so it will be checked by CI

HuStmpHrrr:
could you reply to my reviews if they are not addressed?

theories/Syntax.v
```diff
@@ -35,7 +36,7 @@ with subst : Set :=
   | a_extend : subst -> exp -> subst.
 
 (* Some convenient infix notations *)
-Infix "∘" := a_compose (at level 70).
+(*Infix "∘" := a_compose (at level 70).*)
```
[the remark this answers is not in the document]
IanGrenville: Oh sorry I forgot about this, it was conflicting with notation that was in the standard library stuff I was importing and I couldn't find a way to only enable one of them within Elaborator.v

Would putting the notations in Syntax.v in their own scope fix it?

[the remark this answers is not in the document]
HuStmpHrrr: ok. in that case, we could just simply change the notation, or indeed making it a separate parsing scope would help. but either way, that's out of the scope of this PR. I will create an issue about this.

HuStmpHrrr (approved):

HuStmpHrrr:
Ok I will merge this PR and work on my own proof.

PR #13, opened 2023-03-15
Reorganize Coq modules + AST definition

jinh0:
Since the majority of the code in this project will be written in Coq, I thought we could reorganize the project accordingly.

I was thinking we could have one file `Syntax.v` like https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Syntax.agda that would contain the CST and AST types, so that we could import them in other modules.

Also:
-  It didn't make sense to have a "coq/" folder. Accordingly, the .gitignore and Makefile have been updated.
- "Extract.v" has been renamed to be more detailed: "ParserExtraction.v".
- The name of the Coq module is now "Mcltt" not "McLtt" to coincide with the dune project's name.
- Removed outdated files `ast.ml`, `cst.ml`, and `abstract.ml` (since their counterparts are being written in Coq)


lib/Syntax.v
```diff
@@ -0,0 +1,37 @@
...
+  | App : exp -> exp -> exp
+  (* Substitutions *)
+  | Sub : exp -> subst -> exp
+with subst : Set :=
+  | Id : subst
+  | Weaken : subst
+  | Compose : subst -> subst -> subst
+  | Extend : subst -> exp -> subst.
+
+Definition Ctx : Set := list exp.
```
HuStmpHrrr, line 37: instead of `Definition`, consider making it a `Notation`.

lib/Syntax.v
```diff
@@ -0,0 +1,37 @@
+Require Import String.
+Require Import List.
+
+(* CST term *)
+Module Cst.
+Inductive obj :=
+  | TType : nat -> obj
```
HuStmpHrrr, line 7: maybe avoid capitalize the first letter for constructor. it's not very conventional at least in Coq. it's more common to use `_` with the name (or an abbreviation) of its type as prefix.

lib/Syntax.v
```diff
@@ -0,0 +1,37 @@
...
+  | Var : nat -> exp
+  (* Functions *)
+  | Fun : exp -> exp
+  | App : exp -> exp -> exp
+  (* Substitutions *)
+  | Sub : exp -> subst -> exp
+with subst : Set :=
+  | Id : subst
+  | Weaken : subst
+  | Compose : subst -> subst -> subst
```
HuStmpHrrr, line 34: try to introduce some `Notation` or `Infix`. these will help making the constructors more conventional.

jinh0: I've added infix notations for extension and composition, and I left identity as `id`. I'm not sure what notation weakening should have? In the paper, I couldn't find anything about weakening, unless the up arrow ↑ is weakening.

HuStmpHrrr: No it's fine. This is more like good you too play with the parser. There is no need for a Boston for everything.

HuStmpHrrr: oh god I just read what I wrote on my phone. I hope you got what I meant: there is no need for a notation for everything. I will merge this and when you start with writing down the typing rules, you should find convenient to have some more notations. I will leave you with that in the next PR.

PR #16, opened 2023-03-21
Add typing rules

jinh0:
Could we have a list of the rules this pull request should contain, and their names as well? Right now, I'm going by [the paper's rules/names](https://bitbucket.org/pl-uwr/nbe-martinlof/src/master/Nbe/Syntax/System.v).

P.S. Ignore commits b4339feb6325b134b26fca5d80a1d971ab72cf81 and 336e17d05c14736f18dcd3d0c57ac3da1e3858f4 on the `main` branch. I accidentally pushed my changes to the `main` branch, so I reverted them and put them in this branch. Sorry!

### DIFF
--- a/lib/Syntax.v
+++ b/lib/Syntax.v
@@ -24,7 +24,7 @@ Inductive exp : Set :=
   | typ : nat -> exp
   | var : nat -> exp
   (* Functions *)
-  | fn : exp -> exp
+  | fn : exp -> exp -> exp
   | pi : exp -> exp -> exp
   | app : exp -> exp -> exp
   (* Substitutions *)
--- a/lib/Syntax.v
+++ b/lib/Syntax.v
@@ -25,6 +25,7 @@ Inductive exp : Set :=
   | var : nat -> exp
   (* Functions *)
   | fn : exp -> exp
+  | pi : exp -> exp -> exp
   | app : exp -> exp -> exp
   (* Substitutions *)
   | sub : exp -> subst -> exp
--- a/lib/Syntax.v
+++ b/lib/Syntax.v
@@ -20,7 +20,7 @@ Inductive exp : Set :=
   | zero : exp
   | succ : exp -> exp
   (* Type constructors *)
-  | nat : exp
+  | n : exp
   | typ : nat -> exp
   | var : nat -> exp
   (* Functions *)
--- a/lib/Syntax.v
+++ b/lib/Syntax.v
@@ -39,3 +39,4 @@ Infix "∘" := compose (at level 70).
 Infix "," := extend (at level 80).
 
 Notation Ctx := (list exp).
+Notation Typ := exp.
--- a/lib/System.v
+++ b/lib/System.v
@@ -1,0 +1,22 @@
+Require Import List.
+Import ListNotations.
+
+Require Import Mcltt.Syntax.
+
+Reserved Notation "G |-" (at level 80).
+Reserved Notation "G |- T" (no associativity, at level 80, T at next level).
+Reserved Notation "G |- t : T" (no associativity, at level 80, t at next level).
+
+Inductive wf_ctx : Ctx -> Set :=
+  | empty : [] |-
+  | ext : forall G a,
+      G |- -> (a :: G) |-
+where "G |- " := (wf_ctx G)
+with wf_type : Ctx -> Typ -> Set :=
+  | univ_e : forall G t i,
+      G |- t : (typ i) -> G |- t
+where "G |- t" := (wf_type G t)
+with wf_term : Ctx -> exp -> Typ -> Set :=
+  | univ_unit_f : forall G i,
+      G |- -> G |- nat : (typ i)
+where "G |- t : T" := (wf_term G t T).
--- a/lib/System.v
+++ b/lib/System.v
@@ -10,13 +10,17 @@ Reserved Notation "G |- t : T" (no associativity, at level 80, t at next level).
 Inductive wf_ctx : Ctx -> Set :=
   | empty : [] |-
   | ext : forall G a,
-      G |- -> (a :: G) |-
+      G |- -> a :: G |-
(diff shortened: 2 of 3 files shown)


lib/System.v
```diff
@@ -0,0 +1,22 @@
+Require Import List.
+Import ListNotations.
+
+Require Import Mcltt.Syntax.
+
+Reserved Notation "G |-" (at level 80).
+Reserved Notation "G |- T" (no associativity, at level 80, T at next level).
+Reserved Notation "G |- t : T" (no associativity, at level 80, t at next level).
```
jinh0, line 8: I have no idea what the `at next level` stuff is for but without it, this doesn't work. [For reference](https://bitbucket.org/search?q=repo%3Anbe-martinlof%20Reserved&account=%7B19c5c611-1f6e-4736-ba02-386a36162281%7D).

lib/System.v
```diff
@@ -0,0 +1,22 @@
...
+  | ext : forall G a,
+      G |- -> (a :: G) |-
+where "G |- " := (wf_ctx G)
+with wf_type : Ctx -> Typ -> Set :=
+  | univ_e : forall G t i,
+      G |- t : (typ i) -> G |- t
+where "G |- t" := (wf_type G t)
+with wf_term : Ctx -> exp -> Typ -> Set :=
+  | univ_unit_f : forall G i,
+      G |- -> G |- nat : (typ i)
```
jinh0, line 21: Am I on the right track with these rules so far? I'm referring to both the paper and its source code and yours @HuStmpHrrr , but I am also clueless about how Coq works still, so maybe I'm completely wrong on these.

lib/System.v
HuStmpHrrr: try real `\vdash` in emacs. this will give you unicode. make sure install `company-coq` first. also I suspect that `:` won't work because it is in conflict with `t : T` as a Gallina term. try to find a similar character that is easy to input.

lib/System.v
HuStmpHrrr: you are on the right track, but try to bring a common prefix to these constructors, connected with an underscore.

lib/System.v
```diff
@@ -0,0 +1,22 @@
+Require Import List.
+Import ListNotations.
+
+Require Import Mcltt.Syntax.
+
+Reserved Notation "G |-" (at level 80).
```
HuStmpHrrr, line 6: usually we write `|- G` instead.

lib/System.v
```diff
@@ -0,0 +1,22 @@
...
+Reserved Notation "G |- T" (no associativity, at level 80, T at next level).
```
HuStmpHrrr, line 7: we don't need this judgment. check the agda code.

lib/System.v
HuStmpHrrr: To find out what something in Coq means, use the reference manual: https://coq.inria.fr/distrib/current/refman/index.html#

I didn't know what `at next level` means, but a proper search in the reference manual will help. This is what I did:

1. Since you are asking about an option of `Notation`, I clicked `Indexes` on the left and then chose `Command index`.
2. Then I find `Notation`,
3. On that page, I searched `next level`, which identifies the section: https://coq.inria.fr/distrib/current/refman/user-extensions/syntax-extensions.html#precedences-and-associativity and the sentence
> The level is either a number or the phrase next level whose meaning is obvious.
so next level I guess is the next level that is tighter. I suppose here it means 79. I have not looked deep into the refman but I guess you can test to make it at level `79` and see if it behaves the same. You can learn more about Coq just by reading the refman and I encourage you try to do so.

lib/System.v
jinh0: To clarify, you mean adding a common prefix like, for all the `wf_term` constructors, we'd start them with `term_`, and `wf_ctx` constructors start with `ctx_`?

HuStmpHrrr: maybe not this long, something like `wt_` or `wf_` would suffice. the reason is to introduce some prefix to avoid name clashing that might occur in the future. especially the occasion of a direct reference to a constructor is rare.

lib/System.v
jinh0: From [your Agda code](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Full.agda#L334), I see `G |- T` = there exists `G |- T : Type i`. So am I replacing all `G |- T` with `forall i, G |- T : Type i`?

HuStmpHrrr: I am supposed to remove it. That is not used anywhere.

lib/System.v
```diff
@@ -0,0 +1,24 @@
...
+Reserved Notation "⊢ Γ" (at level 80).
+Reserved Notation "Γ ⊢ t : T" (no associativity, at level 80, t at next level).
+
+Inductive wf_ctx : Ctx -> Set :=
+  | wf_empty : ⊢ []
+  | wf_ext : ∀ Γ a, ⊢ Γ -> ⊢ a :: Γ
+where "⊢ Γ" := (wf_ctx Γ)
+with wf_term : Ctx -> exp -> Typ -> Set :=
+  | wf_univ_unit :
+      ∀ Γ i, ⊢ Γ -> Γ ⊢ n : typ i
```
jinh0, line 16: Are these unicode symbols okay? I don't remember if we said yes to unicode symbols or not.

HuStmpHrrr: yes, it looks good like this.

lib/Syntax.v
```diff
@@ -20,11 +20,12 @@ Inductive exp : Set :=
   | zero : exp
   | succ : exp -> exp
   (* Type constructors *)
-  | nat : exp
+  | n : exp
```
HuStmpHrrr, line 23: this constructor is too short. in general we should have at least 3 characters for the constructors, otherwise it will be too easy to clash with variables in definition and cause unnecessary inconvenience.

PR #21, opened 2023-03-26
Fix parser

### LINKED ISSUE #17: parser is broken
the parser now is broken with the changes made to the syntax file

I've add a CI file. please make sure it actually works.

jinh0:
Closes #17 . Sorry about all this. I forgot the implications of changing the CST term names, I thought it wasn't used anywhere. I should have been more careful.

What has been fixed:
- The `a_extend` notation "," was conflicting with parser compilation
- In addition, for the time being, the names of the CST terms have been reverted.

The parser will compile and you can use it in `utop`. The only caveat is that `test.ml` is completely broken right now due to the weird typing name stuff. I'll fix that in this branch.

From what I understand, the parser has been broken since 50790592232502d1435dbded066b3e5662a4f5b5 since I renamed the CST terms.

### DIFF
--- a/theories/Parser.vy
+++ b/theories/Parser.vy
@@ -24,12 +24,12 @@ prog:
 
 obj:
   | LAMBDA args_list DOT obj {
-      List.fold_left (fun acc arg => Cst.Fun (fst arg) (snd arg) acc) $2 $4 
+      List.fold_left (fun acc arg => Cst.fn (fst arg) (snd arg) acc) $2 $4 
   }
   | PI args_list DOT obj {
-      List.fold_left (fun acc arg => Cst.Pi (fst arg) (snd arg) acc) $2 $4
+      List.fold_left (fun acc arg => Cst.pi (fst arg) (snd arg) acc) $2 $4
   }
-  | SUCC obj { Cst.Succ $2 }
+  | SUCC obj { Cst.succ $2 }
   (* Application is a special case, where we must avoid conflict by associativity: *)
   (* see https://github.com/utgwkk/lambda-chama/blob/master/parser.mly *)
   | app_obj { $1 }
@@ -46,14 +46,14 @@ args_obj:
 (* M N *)
 app_obj:
   (* simpl_obj prevents conflict by associativity *)
-  | app_obj simpl_obj { Cst.App $1 $2 }
+  | app_obj simpl_obj { Cst.app $1 $2 }
   | simpl_obj { $1 }
 
 (* Either a variable or parentheses around a complex object *)
 simpl_obj:
-  | VAR { Cst.Var $1 }
-  | NAT { Cst.Nat }
-  | ZERO { Cst.Zero }
-  | TYPE INT { Cst.TType $2 }
+  | VAR { Cst.var $1 }
+  | NAT { Cst.nat }
+  | ZERO { Cst.zero }
+  | TYPE INT { Cst.typ $2 }
   | LPAREN obj RPAREN { $2 }
 
--- a/theories/Syntax.v
+++ b/theories/Syntax.v
@@ -4,14 +4,14 @@ Require Import List.
 (* CST term *)
 Module Cst.
 Inductive obj : Set :=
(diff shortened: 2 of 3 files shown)


jinh0:
Just a little more detail on why `test.ml` no longer works:

Right now, the extraction spits out this OCaml type:
```ocaml
module Cst =
 struct
  type obj =
  | Coq_typ of int
  | Coq_nat
  | Coq_zero
  | Coq_succ of obj
  | Coq_app of obj * obj
  | Coq_fn of string * obj * obj
  | Coq_pi of string * obj * obj
  | Coq_var of string
 end

```

I am guessing we can just specify in our `ParserExtraction.v` file how to extract this type with the names we want.

theories/Syntax.v
```diff
@@ -4,14 +4,14 @@ Require Import List.
...
 Inductive obj : Set :=
-  | c_typ : nat -> obj
-  | c_nat : obj
-  | c_zero : obj
-  | c_succ : obj -> obj
-  | c_app : obj -> obj -> obj
-  | c_fn : string -> obj -> obj -> obj
-  | c_pi : string -> obj -> obj -> obj
-  | c_var : string -> obj.
+  | typ : nat -> obj
```
HuStmpHrrr, line 7: why roll back `Cst` instead of changing the parser?

HuStmpHrrr: with `c_`, will extraction still put `Coq_` before constructor names?

jinh0: >  why roll back Cst instead of changing the parser?

One thought I had was if we're adding `c_` for namespacing purposes, we could just instead put the type in the module `Cst`: to get the `nat` constructor, you use `Cst.nat` or `Cst.app`, though I do get how that may be less succinct.

> with c_, will extraction still put Coq_ before constructor names?

I get this:
```ocaml
module Cst =
 struct
  type obj =
  | Coq_c_typ of int
  | Coq_c_nat
  | Coq_c_zero
  | Coq_c_succ of obj
  | Coq_c_app of obj * obj
  | Coq_c_fn of string * obj * obj
  | Coq_c_pi of string * obj * obj
  | Coq_c_var of string
 end
```

I was looking into more examples of extraction in [the Extraction folder](https://github.com/coq/coq/blob/master/theories/extraction/ExtrOcamlIntConv.v), but no luck as of yet.

For the time being, since the only thing stopping us from having a working parser is the incorrect constructor names in `test.ml`, I could just change them to the `Coq_` names for the time being.

theories/Parser.vy
jinh0, line 1: The only changes here are making the constructor names lowercased to reflect the lowercase constructors in `Syntax.v`.

theories/Syntax.v
jinh0: You can run them with `dune runtest`. Although right now, `dune build` will throw an error from `test.ml` not because the tests are running but the typing is wrong. I can add running the tests as a part of the CI file.

HuStmpHrrr: yes that will be nice.

jinh0: If the constructor names are capitalized, i.e. `Nat`, then the extracted OCaml type comes out just fine. It's probably adding `Coq_` since lowercase names refer to types not constructors in OCaml.

HuStmpHrrr: it's fine to leave it like this then. we spend most of our time in Coq so it's better to be conventional in Coq.

.github/workflows/ci_build.yml
```diff
@@ -41,3 +41,8 @@ jobs:
           eval $(opam env)
           cd theories/
           make
+      - name: test parser
+        run: |
+          cd ../
```
HuStmpHrrr, line 46: I don't think you should `cd ../` instead, you should `cd` into `driver`? just `cd driver`. also, you will need to do `eval $(opam env)` to have dune visible.

HuStmpHrrr: check and add the installation for the dependencies. also remember to update README

PR #22, opened 2023-04-02
Merge fork into `typing-rules` branch

jinh0:
Closes #3 .

Typing rules to do (in order given by https://bitbucket.org/pl-uwr/nbe-martinlof/src/master/Nbe/Syntax/System.v):

## Context

- [x]  Empty
- [x]  Extension

## Context Equality

- [ ]  Reflexivity
- [ ]  Extension

## Terms

- [x]  Nats
- [ ]  univ_empty
- [x]  univ_unit
- [x]  Function
- [x]  Hypothesis
- [x]  Fun_E
- [x]  Fun_I
- [x]  Zero
- [x]  Succ
- [ ]  Sb
- [ ]  Unit
- [ ]  Conv
- [ ]  Empty_e

## Substitutions

- [x]  Id
- [x]  Weaken
- [x]  Compose
- [x]  Extend
- [ ]  Conv

## Equalities

PR #26, opened 2023-04-29
Add typing rules

### LINKED ISSUE #3: Write down the typing rules for MLTT
Basically repeat these rules:

https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Full.agda

One can use notations to make them look nicer in Coq, c.f. https://coq.github.io/doc/v8.16/refman/user-extensions/syntax-extensions.html#coq:cmd.Notation .

jinh0:
Closes #3 . (This pull request is a continuation of #22 ).

### DIFF
--- a/theories/Syntax.v
+++ b/theories/Syntax.v
@@ -25,8 +25,8 @@ Inductive exp : Set :=
   | a_var : nat -> exp
   (* Functions *)
   | a_fn : exp -> exp -> exp
-  | a_pi : exp -> exp -> exp
   | a_app : exp -> exp -> exp
+  | a_pi : exp -> exp -> exp
   (* Substitutions *)
   | a_sub : exp -> subst -> exp
 with subst : Set :=
@@ -40,3 +40,9 @@ with subst : Set :=
 Infix ",," := a_extend (at level 80).
 
 Notation Ctx := (list exp).
+Notation Sb := subst.
+Notation Typ := exp.
+Notation typ := a_typ.
+Notation ℕ := a_nat.
+Notation 𝝺 := a_fn.
+Notation Π := a_pi.
--- a/theories/Syntax.v
+++ b/theories/Syntax.v
@@ -36,10 +36,9 @@ with subst : Set :=
   | a_extend : subst -> exp -> subst.
 
 (* Some convenient infix notations *)
-Infix "-∘-" := a_compose (at level 70).
+Infix "∙" := a_compose (at level 70).
 Infix ",," := a_extend (at level 80).
 
-Notation "T [ σ ]" := (a_sub T σ) (at level 80, σ at next level).
 
 Notation Ctx := (list exp).
 Notation Sb := subst.
--- a/theories/Syntax.v
+++ b/theories/Syntax.v
@@ -36,8 +36,10 @@ with subst : Set :=
   | a_extend : subst -> exp -> subst.
 
 (* Some convenient infix notations *)
-(*Infix "∘" := a_compose (at level 70).*)
+Infix "-∘-" := a_compose (at level 70).
 Infix ",," := a_extend (at level 80).
+
+Notation "T [ σ ]" := (a_sub T σ) (at level 80, σ at next level).
 
 Notation Ctx := (list exp).
 Notation Sb := subst.
--- a/theories/System.v
+++ b/theories/System.v
@@ -6,9 +6,10 @@ Require Import Mcltt.Syntax.
 
 Reserved Notation "⊢ Γ" (at level 80).
 Reserved Notation "⊢ Γ ≈ Δ" (at level 70).
-Reserved Notation "Γ ⊢ A ≈ B : T" (at level 80).
+Reserved Notation "Γ ⊢ A ≈ B : T" (at level 80, A at next level, B at next level).
 Reserved Notation "Γ ⊢ t : T" (no associativity, at level 80, t at next level).
 Reserved Notation "Γ ⊢ [ e ] : T" (no associativity, at level 80, e at next level).
+Reserved Notation "Γ ⊢s S1 ≈ S2 : Δ" (no associativity, at level 80, S1 at next level, S2 at next level).
 
 Generalizable All Variables.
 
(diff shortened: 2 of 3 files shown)


theories/System.v
```diff
@@ -0,0 +1,95 @@
+Require Import List.
+Require Import Unicode.Utf8_core.
+Import ListNotations.
+
+Require Import Mcltt.Syntax.
+
+Reserved Notation "⊢ Γ" (at level 80).
+Reserved Notation "⊢ Γ ≈ Δ" (at level 70).
+Reserved Notation "Γ ⊢ A ≈ B : T" (at level 80).
```
jinh0, line 9: I'm unable to get these working, something about precedence I think.

HuStmpHrrr:
@IanGrenville could you pick this up and add as many rules as you can. please refer to my agda code as the golden source of information.

IanGrenville:
I've tried to add the remaining rules for our current syntax. There are a few things I wasn't sure about that I'll annotate on here

theories/System.v
```diff
@@ -0,0 +1,241 @@
...
+  | wf_sub : `(
+      Γ ⊢ [s] : Δ ->
+      Δ ⊢ M : A ->
+      Γ ⊢ a_sub M s : a_sub A s
+               )
+  (*| wf_conv : `(
+      Γ ⊢ t : T ->
+      wf_context_eq T T' ->
+      Γ ⊢ t : T'
+      ) *)
```
IanGrenville, line 79: This, and a few other typing rules, matches the Agda rule but I currently have it commented out because when it's uncommented I get an error that "Large non-propositional inductive types must be in Type", and I'm not sure what causes that. From the message I would guess it's about universes, but I'm not sure why this rule would cause a problem with that.

theories/System.v
```diff
@@ -0,0 +1,241 @@
...
+  | wf_eq_pi_cong : `(
+      Γ ⊢ T : typ i ->
+      Γ ⊢ M ≈ M' : typ i ->         
+      M :: Γ ⊢ T ≈ T' : typ i ->              
+      Γ ⊢ Π M T ≈ Π M' T' : typ i              
+    )
+  | wf_eq_var : `(
+      ⊢ a_var i :: Γ ->
+      a_var i :: Γ ⊢ a_var i : T ->
+      a_var i :: Γ ⊢ a_var i ≈ a_var i : T
```
IanGrenville, line 127: This corresponds to a rule in the Agda implementation that uses the $\in !$ predicate, but I saw that previously Jinho had rephrased these in terms of having the variable directly in the context(which it seems is what the CPP 2018 paper does). I went with that approach for this and other instances for now, but can could add the predicate and rephrase these instead if that would be better.

theories/Syntax.v
```diff
@@ -36,7 +36,15 @@ with subst : Set :=
   | a_extend : subst -> exp -> subst.
 
 (* Some convenient infix notations *)
-(*Infix "∘" := a_compose (at level 70).*)
+Infix "-∘-" := a_compose (at level 70).
 Infix ",," := a_extend (at level 80).
 
+Notation "T [ σ ]" := (a_sub T σ) (at level 80, σ at next level).
```
IanGrenville, line 42: I wanted to add this notation to tidy up some of the rules, but it seems to conflict with the [ ] notation for lists. Not sure if there's a way to make them coexist, but it seems like they should given that the $\Gamma \vdash [ \sigma] : \Delta$ notation does work.

IanGrenville:
> > @HuStmpHrrr I've tried to add the remaining rules for our current syntax. There are a few things I wasn't sure about that I'll annotate on here
> 
> how do you annotate? If you do it by review, then you need to make sure you submit them so that they are visible on my end.

Oh sorry, I did forget to hit submit

theories/System.v
HuStmpHrrr: set all these judgments to be in `Prop`.

theories/System.v
HuStmpHrrr: could you give a link to their Coq repo? I don't think this will work. there must be another weakening rule to fix it. please use my version for now.

theories/Syntax.v
HuStmpHrrr: ok. in that case we could use a parsing scope. just remove or comment this notation out for now. postfix `[]` is a frequent notation, which we could probably use a type class for. but let's leave it for the future.

theories/Syntax.v
```diff
@@ -36,7 +36,15 @@ with subst : Set :=
   | a_extend : subst -> exp -> subst.
 
 (* Some convenient infix notations *)
-(*Infix "∘" := a_compose (at level 70).*)
+Infix "-∘-" := a_compose (at level 70).
```
HuStmpHrrr, line 39: this notation is a bit counter-intuitive. are you able to find another dot that looks similar?

theories/System.v
```diff
@@ -0,0 +1,241 @@
...
+Inductive wf_ctx : Ctx -> Set :=
+  | wf_empty : ⊢ []
+  | wf_extend : `(
+      ⊢ Γ ->
+      Γ ⊢ T : typ i ->
+      ⊢ T :: Γ
+    ) 
+where "⊢ Γ" := (wf_ctx Γ)
+with wf_ctx_eq : Ctx -> Ctx -> Set :=
+  | wfc_empty : wf_ctx_eq [] []
```
HuStmpHrrr, line 25: maybe replace all `wf_ctx_eq` with the notation?

theories/System.v
IanGrenville: This is what I was thinking of in their repo:
https://bitbucket.org/pl-uwr/nbe-martinlof/src/25c5be555f09dcb20d2819105e0c46981c65bb96/Nbe/Syntax/System.v#lines-99 

I added the predicate and changed the relevant rules.

theories/Syntax.v
IanGrenville: I'll try using $\bullet$ instead.

HuStmpHrrr:
is it ready to merge?

HuStmpHrrr (approved):

HuStmpHrrr:
never mind. i am merging.

PR #28, opened 2023-05-09
Add lemmas for context conversion proof

IanGrenville:
- [x] [presup-⊢≈](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Properties/Contexts.agda#L28)
- [x] [∈!⇒ty-wf](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Properties/Contexts.agda#L18)
- [x] [∈!⇒ty≈](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Properties/Contexts.agda#L36)
- [x] [[]-cong-Se′](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Misc.agda#L57)
- [x] [⊢≈-sym](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Properties/Contexts.agda#L49)
- [x] [≈-refl](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Refl.agda#L10)
- [x] [t[σ]-Se](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Misc.agda#L51)
- [x] Extracting information from $\vdash T :: \Gamma$


IanGrenville:
I decided I would also need to port the [presup lemmas](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Presup.agda) to prove the other lemmas. 

The problem is that in the Agda implementation, they're implemented as a mutual induction proof, which seems harder to do in Coq. I tried doing it with one big Lemma statement that includes each sublemma in a `with` clause, but ran into issues because doing the proof that way doesn't check the validity of my induction. So for now I've just tried proving each lemma given the assumption of the others when neccesary, and hopefully it shouldn't be too hard to assemble those into a proper mutual inductive proof.

theories/PresupLemmas.v
```diff
@@ -0,0 +1,480 @@
...
+Lemma ctx_decomp (Γ : Ctx) (T : Typ) : ⊢ T :: Γ -> (⊢ Γ ∧ ∃ i, Γ ⊢ T : typ i).
+Proof.
+  intros.
+  destruct H.
+  (* Using either destruct or induction on H forces us to consider a case where H : ⊢ nil, which doesn't make sense
+     and so we can't really prove anything for it. *)
+  admit.
+  (* If we then skip that problematic case, the information we get from destructing H doesn't actually apply
+     because it refers to fresh instances of Γ and T for some reason *)
+Admitted.
```
IanGrenville, line 15: @HuStmpHrrr 

I could use help with this lemma. In the Agda code this seems easily done by pattern matching, like the `⊢TΓ@(⊢∷ ⊢Γ ⊢T)` used [here](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Properties/Contexts.agda#L23), but I can't figure out how to do it the same sort of thing in Coq.

HuStmpHrrr:
> I decided I would also need to port the [presup lemmas](https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Presup.agda) to prove the other lemmas.
> 
> The problem is that in the Agda implementation, they're implemented as a mutual induction proof, which seems harder to do in Coq. I tried doing it with one big Lemma statement that includes each sublemma in a `with` clause, but ran into issues because doing the proof that way doesn't check the validity of my induction. So for now I've just tried proving each lemma given the assumption of the others when neccesary, and hopefully it shouldn't be too hard to assemble those into a proper mutual inductive proof.

I am surprised! `with` should just work! I'd imagine you just need to be a bit more careful. do you have a commit to replay the problem you run into? there are a few quirks in Coq to do mutual induction but all of them can be overcome. `with` is my favorite solution. `Scheme` command is also a solution but just a bit more elaborate: https://coq.inria.fr/refman/proofs/writing-proofs/reasoning-inductives.html#coq:cmd.Scheme

HuStmpHrrr:
I don't think starting from presupposition will work because it depends on context conversion lemmas.

theories/PresupLemmas.v
HuStmpHrrr: use `inversion` instead: https://coq.inria.fr/refman/proofs/writing-proofs/reasoning-inductives.html#coq:tacn.inversion

`dependent destruction` is also an option: https://coq.inria.fr/refman/proofs/writing-proofs/reasoning-inductives.html#coq:tacn.dependent-destruction

there are many useful tactics in the refman. I suggest you to first take a pause and glance what's available in Coq first, so that next time you will at least have an impression of what Coq can do. I will review your code after I get off.

theories/CtxEqLemmas.v
```diff
@@ -0,0 +1,92 @@
...
+Require Import Mcltt.PresupLemmas.
+Require Export Lia.
+
+(* Corresponds to t[σ]-Se in the Agda proof *)
+Lemma sub_lvl (Δ Γ : Ctx) (T : Typ) (σ : Sb) (i : nat) : (Δ ⊢ T : typ i) -> (Γ ⊢s σ : Δ) -> (Γ ⊢ a_sub T σ : typ i).
+Proof.
+  intros.
+  pose proof (wf_sub _ _ _ _ _ H0 H).
+  apply (wf_conv _ _ _ _ (i+1) H1).
+  exact (wf_eq_typ_sub _ _ _ _ H0).
```
HuStmpHrrr, line 13: `exact` is too low-level. I probably will show you how to write proofs without being so elaborate.

HuStmpHrrr: tomorrow

IanGrenville:
> I don't think starting from presupposition will work because it depends on context conversion lemmas.

Dang, I missed that! That means I'll need to rethink how I prove the other the lemmas. At least I'll have a good chunk of the work done once the context conversion stuff is done.

HuStmpHrrr:
> > I don't think starting from presupposition will work because it depends on context conversion lemmas.
> 
> Dang, I missed that! That means I'll need to rethink how I prove the other the lemmas. At least I'll have a good chunk of the work done once the context conversion stuff is done.

don't rush to spend more time on proving now. I think the next step for you is to learn more tactics so that you pick the right tactics at the right time. now your proof scripts, frankly, is too elementary to a point that it negatively impacts your efficiency. 

I would suggest you read a few pages in the refman first:

1. the tactic page I pasted above;
2. read about hint databases in Coq. In particular, I want you to understand `Hint Resolve` and `Hint Constructors` and their relation to hint databases: https://coq.inria.fr/distrib/V8.16.1/refman/proofs/automatic-tactics/auto.html#coq:cmd.Hint-Resolve and https://coq.inria.fr/distrib/V8.16.1/refman/proofs/automatic-tactics/auto.html#coq:cmd.Hint-Constructors
3. I've started a tactic library; https://github.com/Beluga-lang/McLTT/blob/main/theories/LibTactics.v#L5 . use the hint database I created.
4. as a future plan, I also want you to understand `Hint Rewrite`: https://coq.inria.fr/distrib/V8.16.1/refman/proofs/automatic-tactics/auto.html#coq:cmd.Hint-Rewrite this is a significant work horse which does not exist in agda and I miss a lot.

let's catch up tomorrow. before the meeting, it's better for you to read these documents instead of working on the proofs.

IanGrenville:
@HuStmpHrrr 

I've completed proofs for all these lemmas, and have some partial progress on the context equivalence proofs that I'll put up after this is merged.

HuStmpHrrr:
CI fails. maybe check if you have the right Coq version if it works for you locally?

HuStmpHrrr:
but otherwise, thanks. the proof script looks very concise now. I can definitely see that you are building up some expertise.

IanGrenville:
> CI fails. maybe check if you have the right Coq version if it works for you locally?

It looks like it broke when I put the hints in our database instead of the core one. I just put it back for now. Is there a way to make it so auto uses our hint DB without using `auto with mcltt` every time?

HuStmpHrrr:
> > CI fails. maybe check if you have the right Coq version if it works for you locally?
> 
> It looks like it broke when I put the hints in our database instead of the core one. I just put it back for now. Is there a way to make it so auto uses our hint DB without using `auto with mcltt` every time?

write a quick `Ltac` that does that. it is common that have multiple hint databases to interplay with each other.

IanGrenville:
I ended up doing it as a notation so I could still preserve the ability to do `eauto using ...`, but it seems to work fine now.

HuStmpHrrr:
ah, I left some comments in the wrong link: https://github.com/Beluga-lang/McLTT/commit/9ca876d9d021e88548df32a3c175e5f5738c2e13#diff-5e44045da89fc651a84aa144ab189c6f5e72dd96223cbc9ac04da57a62ffc264R43

theories/CtxEqLemmas.v
```diff
@@ -0,0 +1,145 @@
...
+  intros.
+  induction H0.
+  - inversion H.
+    mauto using sub_lvl.
+  - inversion H.
+    destruct (IHctx_lookup H3).
+    eauto using wf_sb_weaken, sub_lvl with mcltt.
+Qed.
+
+Lemma presup_sb_ctx (Γ Δ: Ctx) (σ : Sb) : Γ ⊢s σ : Δ -> ⊢ Γ ∧ ⊢ Δ.
```
HuStmpHrrr, line 75: is this lemma and the one below particularly useful? presupposition will force you to prove them again,

HuStmpHrrr (approved):

PR #36, opened 2023-06-21
Define the NbE algorithm and the PER model

IanGrenville:
Still a work in progress in the PER aspect. The stuff for the NbE algorithm should be good I think, but I may have missed something. I'm going to note the spots with the PER definition that I'm not sure about.

### DIFF
--- a/theories/PER.v
+++ b/theories/PER.v
@@ -16,7 +16,7 @@ Generalizable All Variables.
 
 Notation "a ≈ b ∈ P" := (P a b) (at level 80). 
 Notation "a ∈' P" := (P a a) (at level 80).
-Notation "a ~ b ∈ P" := (P a b) (at level 80).
+
 
 Definition Ty : Type := relation D.
 Definition Ev : Type := relation Env.
@@ -107,7 +107,6 @@ Equations per_U_wf (i : nat) : Ty by wf (i) (lt) :=
   per_U_wf (i) := per_U (i) (λ k p, per_U_wf k)
 .
 
-Definition U_PER (i : nat) : Ty := per_U i (λ k p, per_U_wf k). 
 
 Definition UniInterp (i : nat) : D -> D -> Ty -> Prop := InterpUniv i (λ k p, per_U_wf k).
 
@@ -117,7 +116,7 @@ Record RelTyp (i : nat) (T T' : exp) (p p' : Env) : Set := mk_rel_typ
   ; val_T' : D
   ; eval_T : ⟦ T ⟧ p ↘ val_T
   ; eval_T' : ⟦ T' ⟧ p' ↘ val_T'
-  ; eq_TT' : val_T ≈ val_T' ∈ U_PER i
+  ; eq_TT' : val_T ≈ val_T' ∈ per_U_wf i
   }.
 
 Inductive ctx_equiv_PER : Ctx -> Ctx -> Prop :=
@@ -139,9 +138,9 @@ with InterpCtx : Ctx -> Ctx -> Ev -> Prop :=
                  InterpCtx (T :: Γ) (T' :: Δ)
                    (
                      λ (p p' : Env),
+                       (d_drop p) ≈ (d_drop p') ∈ R ∧
                      ∀ (rel_typ : RelTyp i T T' p p'),
-                     (d_drop p) ≈ (d_drop p') ∈ R ∧
-                      UniInterp i (val_T _ _ _ _ _ rel_typ) (val_T' _ _ _ _ _ rel_typ) P ->
+                       UniInterp i (val_T _ _ _ _ _ rel_typ) (val_T' _ _ _ _ _ rel_typ) P ->
                      (d_lookup p 0) ≈ (d_lookup p' 0) ∈ P
                    )
                  ).
--- a/theories/PER.v
+++ b/theories/PER.v
@@ -76,18 +76,19 @@ Section PERDef.
                de ≈ de' ∈ Bot  ->
                InterpUniv (↑ (d_typ j) de) (↑ (d_typ j) de') per_neu)
   | iu_nat : InterpUniv d_nat d_nat per_nat
-  | iu_pi : `(
-             InterpUniv a b U_ab ->
-             a ≈ b ∈ U_ab ->
-             pi_RT T T' (p ↦ a) (p' ↦ b) U_im ->
-             InterpUniv (d_pi a T p) (d_pi b T' p') (λ f f',
-                 ∃ fa fb, (f ∙d a ↘ fa) ∧ (f' ∙d b ↘ fb) ∧ (fa ≈ fb ∈ U_im)
-               )
-              ) 
+  | iu_pi : `( A ≈ A' ∈ per_U ->
+               InterpUniv A A' per_A ->
+               (∀ a a', a ≈ a' ∈ per_A ->
+                        pi_RT T T' (p ↦ a) (p' ↦ a') U_im
+               ) ->
+               InterpUniv (d_pi A T p) (d_pi A' T' p') (λ f f',
+                   ∃ fa fa' : D , ∀ a a', a ≈ a' ∈ per_A ->
+                    (f ∙d a ↘ fa) ∧ (f' ∙d b ↘ fb) ∧ (fa ≈ fb ∈ U_im)                  
+                 )
+               )  
   | iu_univ : (∀ j (p : j < i),
                 InterpUniv (d_typ j) (d_typ j) (Univ j (p)))
   .
-  (* Need another case for universe, maybe needs to be mutual with per_U *)
 
 
   Inductive per_U : Ty :=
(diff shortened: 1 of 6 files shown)


theories/PER.v
```diff
@@ -0,0 +1,142 @@
...
+Require Import Classes.RelationClasses.
+Require Import Unicode.Utf8.
+
+Generalizable All Variables.
+
+Notation "a ≈ b ∈ P" := (P a b) (at level 80). 
+Notation "a ∈' P" := (P a a) (at level 80).
+Notation "a ~ b ∈ P" := (P a b) (at level 80).
+
+Definition Ty : Type := relation D.
```
HuStmpHrrr, line 17: unlike Agda, every definition in Coq induces a level of indirection during proving. usually there are two ways to work around that: either use Notation instead of Definition, or use the Arguments command to ask simpl to unfold the definitions.

theories/Syntax.v
```diff
@@ -47,3 +60,23 @@ Notation typ := a_typ.
...
+  | ne_rec T z s u => a_rec (NfToExp T) (NfToExp z) (NfToExp s) (NeToExp u) 
+  | ne_app u n=> a_app (NeToExp u) (NfToExp n)
+  end
+with NfToExp (nf : Nf) : exp :=
+  match nf with
+  | nf_ne u => NeToExp u
+  | nf_zero => a_zero
+  | nf_succ w => a_succ (NfToExp w)
+                   (* !!Wrong!! *)    
+  | nf_lam w => a_fn (NfToExp w)(NfToExp w)
```
IanGrenville, line 77: I wasn't sure what to do here, since in the Agda proof nf_lam only takes one argument, but our usual lambda term needs two, so when converting I don't know what to put for the other argument

IanGrenville (commented):
Sorry, I forgot to press the button to post my comments again @HuStmpHrrr

theories/Syntax.v
HuStmpHrrr: this is correct. conversion from normal form to expression has no catchya.

theories/PER.v
```diff
@@ -0,0 +1,142 @@
...
+  | per_pi : `(DA ≈ DA ∈ per_U ->
+               InterpUniv DA PA ->
+               (∀ a, a ≈ a ∈ PA -> ∃ DB, DF ∙d a ↘ DB) ->
+               (∀ a, a ≈ a ∈ PA -> ∃ DB', DF' ∙d a ↘ DB') ->
+               (InterpUniv DA P -> a0 ≈ a1 ∈ P ->
+                 DF ∙d a0 ↘ DB0 ->
+                 DF' ∙d a1 ↘ DB1 ->
+                 DB0 ≈ DB1 ∈ per_U) ->
+               d_pi DA DF ≈ d_pi DA' DF' ∈ per_U)
+  .
```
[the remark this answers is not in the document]
HuStmpHrrr: I guess we need to discuss it in the meeting.

theories/PER.v
```diff
@@ -0,0 +1,142 @@
...
+
+ 
+  Inductive InterpUniv : D -> Ty -> Prop :=
+  | iu_ne : `( j < i ->
+               de ≈ de ∈ Bot  ->
+               InterpUniv (↑ (d_typ j) de) per_neu)
+  | iu_nat : InterpUniv d_nat per_nat
+  | iu_pi : `(InterpUniv DA PA ->
+              ProperPerProd PA PF ->
+              (∀ a PB DB, a ≈ a ∈ PA -> DF ∙d a ↘ DB -> PF a PB -> InterpUniv DB PB) ->
```
HuStmpHrrr, line 122: this kinda predicate doesn't really look right.

theories/PER.v
```diff
@@ -0,0 +1,116 @@
...
+  | per_u_ne : `(C ≈ C' ∈ Bot -> ↑ A C ≈ ↑ A C' ∈ per_U)
+  | per_u_nat : `(d_nat ≈ d_nat ∈ per_U)
+  | per_u_univ : `(j < i -> j = j' -> d_typ j ≈ d_typ j' ∈ per_U)
+  | per_u_pi : `(
+               A ≈ A' ∈ per_U ->
+               InterpUniv A A' per_A ->
+               a ≈ a' ∈ per_A ->
+               pi_RT T T' (p ↦ a) (p' ↦ a') per_U ->
+               d_pi A T p ≈ d_pi A' T' p' ∈ per_U
+               )
```
IanGrenville, line 103: I think this definition should be right, it should be saying that if the input types are equivalent and we have equivalent inputs, then the pi types are equivalent if the inputs evaluate to equivalent results.

theories/PER.v
```diff
@@ -0,0 +1,116 @@
...
+               )   
+  . 
+End PERDef.
+
+Check lt.
+
+Equations? per_U_wf (i j : nat) : j < i -> Ty by wf (j) (lt) :=
+  per_U_wf (n) (m) (p) := per_U m (λ k p', per_U_wf (n) k (lt_trans k m n  p' p)).
+Proof.
+Admitted.
```
IanGrenville, line 112: I tried to use the Coq equations module to define this like we talked about, but it throws a mysterious error when I try and define it normally. 
`Error: <in exception printer>:<original exception:Anomaly "Uncaught exception Not_found."`
The error showed up when I put in the "by wf" part. Before that it just complained about not being able to find the fixpoint. But weirdly, when i put it in the proof mode for equations it works and says there are no goals/obligations to deal with, but marking it QED still causes an error, but if we admit it it can't be used.

theories/PER.v
```diff
@@ -0,0 +1,116 @@
...
+               de ≈ de' ∈ Bot  ->
+               InterpUniv (↑ (d_typ j) de) (↑ (d_typ j) de') per_neu)
+  | iu_nat : InterpUniv d_nat d_nat per_nat
+  | iu_pi : `(
+             InterpUniv a b U_ab ->
+             a ≈ b ∈ U_ab ->
+             pi_RT T T' (p ↦ a) (p' ↦ b) U_im ->
+             InterpUniv (d_pi a T p) (d_pi b T' p') (λ f f',
+                 ∃ fa fb, (f ∙d a ↘ fa) ∧ (f' ∙d b ↘ fb) ∧ (fa ≈ fb ∈ U_im)
+               )
```
IanGrenville, line 85: This seems right to me for the relation we want to interpret pi types to have, as far as I can tell it corresponds to the definition we have in the universe PER, but I'm not fully sure. Do you think it makes sense?

IanGrenville:
@HuStmpHrrr I think I have the right definitions for the PER, but I'm not fully sure. Would be great if you could take a look. I also ran into an issue with coq-equations for the next step that I'm not sure how to resolve

theories/PER.v
HuStmpHrrr: this interpretation is incorrect. universal quantifications are missing.

IanGrenville:
@HuStmpHrrr I changed the definition like we talked about on Thursday, and I also managed to get the Equation definition for the universe PER being wellfounded working. I hope the definition is right now. 

The last thing I'm not sure about is how to port the [PER model for contexts](https://gitlab.com/JasonHuZS/practice/-/blame/master/MLTT/Semantics/PER.agda#L164) without induction-recursion. I took a look at what the CPP18 implementation does, they once again use an inductively defined proposition, but their definition of environments is pretty different, so I'm not sure trying to port it will be helpful. 

What I'd like to do is define 
`Inductive ctx_equiv_PER : Ctx -> Ctx -> Prop :=`

  ...

 ` with InterpCtx : ctx_equiv_PER -> Ev -> Prop := `
  ...

But Coq won't seem to allow the mutually defined types to reference the other in its signature like that. I've currently been trying to define InterpCtx as `Ctx -> Ctx -> Ev -> Prop`, but then I'm not sure how to actually define in it such a way that I can get the relevant information about the contexts and their relationship.

theories/PER.v
```diff
@@ -0,0 +1,158 @@
...
+                     UniInterp i d d' P ->
+                     (d_lookup p 0) ≈ (d_lookup p' 0) ∈ P
+                   )
+                   (*
+                     λ (p p' : Env),
+                     ∀ (rel_typ : RelTyp i T T' p p'),
+                     (d_drop p) ≈ (d_drop p') ∈ R ->
+                     RelTyp i T T' p p' ->
+                     UniInterp i (val_T _ _ _ _ _ rel_typ) (val_T' _ _ _ _ _ rel_typ) P ->
+                     (d_lookup p 0) ≈ (d_lookup p' 0) ∈ P
```
IanGrenville, line 154: I came up with two different definitions here because I wasn't sure about the right way to go about it. I feel like it makes sense to make use of the RelTyp record, but I couldn't figure out a way to have it as one of the premises in a way that seems right and also allows me to access it arguments, so I left it commented out and put a more direct definition instead.

theories/PER.v
```diff
@@ -0,0 +1,158 @@
...
+    )
+with InterpCtx : Ctx -> Ctx -> Ev -> Prop :=
+| InterpEmp : InterpCtx nil nil (λ p p', True)
+| InterpCong :
+               `(ctx_equiv_PER Γ Δ ->
+                 InterpCtx Γ Δ R ->
+                 InterpCtx (T :: Γ) (T' :: Δ)
+                   (
+                     λ (p p' : Env),
+                     (d_drop p) ≈ (d_drop p') ∈ R ->
```
HuStmpHrrr, line 142: you are getting close. this line and below should be conjunctive. you'd better wrap things up properly using records. writing things out completely like this is likely confusing you. the same as `iu_pi` above, use a proper definition.

theories/PER.v
HuStmpHrrr: i see. then I suppose some engineering is needed. you can't just take my agda code for granted because there is a difference between `Set` and `Prop`.  anyways, let's keep it like this for now.

theories/PER.v
```diff
@@ -0,0 +1,158 @@
...
+                     ⟦ T ⟧ p ↘ d ->
```
HuStmpHrrr, line 143: these `d`s will be universally quantified by `` `( ``!

IanGrenville:
Okay I changed the definition to the one you said to go with and tweaked it a bit. I think it seems right, although I'm not fully sure if the RelTyp term should be existentially or universally quantified? 

I was also wondering the same about the corresponding ctx_cong definition. Would it be possible that we could have one set of related environments in which T and T' evaluate to related values, while having other pairs of environments where they don't?

theories/PER.v
```diff
@@ -0,0 +1,149 @@
...
+    )
+with InterpCtx : Ctx -> Ctx -> Ev -> Prop :=
+| InterpEmp : InterpCtx nil nil (λ p p', True)
+| InterpCong :
+               `(ctx_equiv_PER Γ Δ ->
+                 InterpCtx Γ Δ R ->
+                 InterpCtx (T :: Γ) (T' :: Δ)
+                   (
+                     λ (p p' : Env),
+                     ∀ (rel_typ : RelTyp i T T' p p'),
```
HuStmpHrrr, line 142: this universal is correct, but it should be easy to fill in. I think that the `(∀ p p',
          p ≈ p' ∈ R ->
          RelTyp i T T' p p' 
      )` premise should also come in here.

theories/PER.v
```diff
@@ -0,0 +1,149 @@
...
+               d_pi A T p ≈ d_pi A' T' p' ∈ per_U
+               ))   
+  . 
+End PERDef.
+
+Equations per_U_wf (i : nat) : Ty by wf (i) (lt) :=
+  per_U_wf (i) := per_U (i) (λ k p, per_U_wf k)
+.
+
+Definition U_PER (i : nat) : Ty := per_U i (λ k p, per_U_wf k).
```
HuStmpHrrr, line 110: you most definitely don't need this definition. if you compare this definition and `per_U_wf`, you should see the bodies are identical.

theories/PER.v
```diff
@@ -0,0 +1,149 @@
...
+with InterpCtx : Ctx -> Ctx -> Ev -> Prop :=
+| InterpEmp : InterpCtx nil nil (λ p p', True)
+| InterpCong :
+               `(ctx_equiv_PER Γ Δ ->
+                 InterpCtx Γ Δ R ->
+                 InterpCtx (T :: Γ) (T' :: Δ)
+                   (
+                     λ (p p' : Env),
+                     ∀ (rel_typ : RelTyp i T T' p p'),
+                     (d_drop p) ≈ (d_drop p') ∈ R ∧
```
HuStmpHrrr, line 143: this conjunction has nothing to do with `rel_typ` and therefore should be promoted outside of the quantification.

IanGrenville:
@HuStmpHrrr Is there anything further I should add? I've noticed there's an issue where the build is failing but it's working on my end. I think it's because I have a more up to date version of Coq or Coq-equations, but I'm not sure.

HuStmpHrrr:
Either update the CI or use the same version for everything as the CI

PR #40, opened 2023-12-15
Update project structure

Ailrun:
This PR puts the static parts into `Core/Syntactic` (and for future, we can put others into `Core/Semantic`, `Core/Completeness`, and `Core/Soundness`) and the more front-end-related parts into `Frontend`.


HuStmpHrrr:
any idea why the CI is failing?

Ailrun:
No idea... it worked in my machine, so failing from the import line is quite weird to me.

Ailrun:
Actually, now I find the reason. It was because of the committed `Parser.v`. I will fix `.gitignore` file to make it work better.

PR #44, opened 2024-01-16
Add semantic

Ailrun:


Ailrun:
Please check whether these definitions make sense;

ref: http://www.cs.nott.ac.uk/~pszvc/publications/induction_recursion.pdf

HuStmpHrrr:
Thanks. There are modifications touching the syntactic files. Maybe they should go to another PR?

HuStmpHrrr:
I'm not sure which way works better. I was thinking about defining the PER model inductively in one shot. Basically, we define the PER relation for types, and then this relation is a functional relation that also returns the relation relating the values. Then we only need one definition, though it might be somewhat complex.

Ailrun:
Providing a functional relation for the PER of values has an issue with Pi case. If we use `Prop` as the type of this record
```Coq
Record rel_mod_app_R R f a f' a' := mk_rel_mod_app_R
  { rel_mod_app_R_fa : domain
  ; rel_mod_app_R_f'a' : domain
  ; rel_mod_app_R_app_fa : {{ $| f & a |↘ rel_mod_app_R_fa }}
  ; rel_mod_app_R_app_f'a' : {{ $| f' & a'|↘ rel_mod_app_R_f'a' }}
  ; rel_mod_app_R_eq : {{ Dom rel_mod_app_R_fa ≈ rel_mod_app_R_f'a' ∈ R }}
  }
.
```
(or with an equivalent formulation using `exist`) then we cannot access `rel_mod_app_R_fa` or `rel_mod_app_R_f'a'`.
However, if we do not use `Prop` for this, we cannot give
```Coq
Inductive per_elem_graph : domain -> domain -> relation domain -> Prop :=
| per_elem_graph_pi :
  (* ... *)
    per_elem_graph d{{{ Π a p B }}} d{{{ Π a' p' B' }}}
      (fun f f' => forall {c c'},
        (* ... *)
        rel_mod_app_R elem_rel' f c f' c') }

```
which eventually returns a non-`Prop` type, thus invalid for the `per_elem_graph` definition.

Ailrun:
But Now I realize that they should be `Prop`, as we have something like https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Semantics/Properties/PER.agda#L443 i.e. `rel_*` that depends on another thing of type `Prop`.

Ailrun:
Actually by changing fields of `rel_mod_eval_R` a bit I think I can restore the projection functions. Then any approach (this one or something like #36) would work.

Ailrun:
Actually by changing fields of `rel_mod_eval_R` a bit I think I can restore the projection functions. Then any approach (this one or something like #36) would work.

HuStmpHrrr:
Yeah I don't see any problem doing what I proposed, other than some back and forth to get the definition right. You need to define a functional relation for the values anyways, so essentially there is no need for separate inductive definitions.

I think this PR is getting big and getting the definition right might take time, while the definitions for operations are fixed. I think this PR should be split into 2 or 3 to get some low hanging fruits merged first.

theories/Core/Semantic/PER.v
```diff
@@ -0,0 +1,183 @@
...
+
+Generalizable All Variables.
+
+Record rel_mod_eval (R : relation domain) A p A' p' : Prop := mk_rel_mod_eval
+  { rel_mod_eval_A : exists a, {{ ⟦ A ⟧ p ↘ a }}
+  ; rel_mod_eval_A' : exists a', {{ ⟦ A' ⟧ p' ↘ a' }}
+  ; rel_mod_eval_equiv : forall {a a'}, {{ ⟦ A ⟧ p ↘ a }} -> {{ ⟦ A' ⟧ p' ↘ a' }} -> {{ Dom a ≈ a' ∈ R }}
+  }
+.
+Arguments mk_rel_mod_eval {_} {_} {_} {_} {_}.
```
HuStmpHrrr, line 17: All these `Arguments` are not necessary. use `Set Implicit Arguments.` c.f. https://coq.inria.fr/refman/language/extensions/implicit-arguments.html#coq:flag.Implicit-Arguments

theories/Core/Semantic/Domain.v
```diff
@@ -0,0 +1,55 @@
...
+Notation "'zero'" := d_zero (in custom domain at level 0) : exp_scope.
+Notation "'succ' m" := (d_succ m) (in custom domain at level 30, m custom domain at level 30) : exp_scope.
+Notation "'rec' m 'under' p 'return' P | 'zero' -> mz | 'succ' -> MS 'end'" := (d_natrec p P mz MS m) (in custom domain at level 0, P custom exp at level 60, mz custom domain at level 60, MS custom exp at level 60, p custom domain at level 60, m custom domain at level 60) : exp_scope.
+Notation "'!' n" := (d_var n) (in custom domain at level 0, n constr at level 0) : exp_scope.
+Notation "'⇑' a m" := (d_neut a m) (in custom domain at level 0, a custom domain at level 30, m custom domain at level 30) : exp_scope.
+Notation "'⇓' a m" := (d_dom a m) (in custom domain at level 0, a custom domain at level 30, m custom domain at level 30) : exp_scope.
+Notation "'⇑!' a n" := (d_neut a (d_var n)) (in custom domain at level 0, a custom domain at level 30, n constr at level 0) : exp_scope.
+
+Definition empty_env : env := fun x => d{{{ zero }}}.
+Definition extend_env (p : env) (d : domain) : env :=
```
HuStmpHrrr, line 46: This definition will cause Coq to reduce `extend_env` to a lambda once two arguments are provided. Maybe use `Arguments` to make it reduce only when there are two?

theories/Core/Syntactic/CtxEquiv.v
```diff
@@ -1,27 +1,26 @@
...
+From Mcltt Require Import SystemLemmas.
 
 #[local]
 Ltac gen_ctxeq_helper_IH ctxeq_exp_helper ctxeq_exp_eq_helper ctxeq_sub_helper ctxeq_sub_eq_helper H :=
   match type of H with
-  | {{ ?Γ ⊢ ?M : ?A }} => pose proof ctxeq_exp_helper _ _ _ H
-  | {{ ?Γ ⊢ ?M ≈ ?N : ?A }} => pose proof ctxeq_exp_eq_helper _ _ _ _ H
-  | {{ ?Γ ⊢s ?σ : ?Δ }} => pose proof ctxeq_sub_helper _ _ _ H
-  | {{ ?Γ ⊢s ?σ ≈ ?τ : ?Δ }} => pose proof ctxeq_sub_eq_helper _ _ _ _ H
+  | {{ ~?Γ ⊢ ~?M : ~?A }} => pose proof ctxeq_exp_helper _ _ _ H
```
HuStmpHrrr, line 10: What's the `~` here?

theories/Frontend/Elaborator.v
```diff
@@ -102,7 +101,7 @@ Inductive closed_at : exp -> nat -> Prop :=
 Hint Constructors closed_at: core.
 
 (*Lemma for the well_scoped proof, lookup succeeds if var is in context*)
-Lemma lookup_known (s : string) (ctx : list string) (H_in : List.In s ctx) : exists n : nat, (lookup s ctx = Some n /\ n < length ctx).
+Lemma lookup_known (s : string) (ctx : list string) (H_in : List.In s ctx) : exists n : nat, (lookup s ctx = Some n /\ n < List.length ctx).
```
HuStmpHrrr, line 104: What was changed to make a quantified path required?

theories/Core/Syntactic/CtxEquiv.v
Ailrun: Splicing-in syntax for the notation

theories/Frontend/Elaborator.v
Ailrun: The order of import. But as there's two `length` in the scope anyway, it is a good idea to use a qualified identifier here.

Ailrun:
> this PR is too long to review. I think you should split it and use one PR exclusively for discussion about the PER model.

Sure, once #45 is finished, I will put another for domain/eval/readback definition, and then this PR will only have PER definition.

Ailrun:
Actually, I will close this and open a new PR that is solely for PER.

PR #49, opened 2024-05-02
Finish PER transitivity

Ailrun:


theories/Core/Semantic/PERLemmas.v
```diff
@@ -118,6 +118,22 @@ Proof.
   intros. eauto using per_univ_elem_right_irrel_gen.
 Qed.
 
+Ltac per_univ_elem_right_irrel_rewrite :=
```
HuStmpHrrr, line 121: What do you clear one but not the other? I suppose you can always clear a duplicated assumption?

theories/Core/Semantic/PERLemmas.v
```diff
@@ -162,24 +178,100 @@ Proof.
...
+  repeat match goal with
+    | H1 : {{ DF ~?A ≈ ~?B ∈ per_univ_elem ?i ↘ ?R1 }}, H2 : {{ DF ~?A' ≈ ~?B ∈ per_univ_elem ?i ↘ ?R1 }} |- _ =>
+        mark H2
+    | H1 : {{ DF ~?A ≈ ~?B ∈ per_univ_elem ?i ↘ ?R1 }}, H2 : {{ DF ~?A' ≈ ~?B ∈ per_univ_elem ?i ↘ ?R2 }} |- _ =>
+        assert (R2 = R1) by (eapply per_univ_elem_left_irrel; eassumption); subst; mark H2
+    end; unmark_all.
+
+Ltac per_univ_elem_irrel_rewrite := functional_per_univ_elem_rewrite_clear; per_univ_elem_right_irrel_rewrite; per_univ_elem_left_irrel_rewrite.
+
+Lemma per_univ_elem_trans : forall i A1 A2 R1,
```
HuStmpHrrr, line 205: I see you are finished. Anything else you still want to do? I think I will try my idea to get rid of indefinite description. Then all these existential conclusions will become unnecessary, but we can merge this PR for now if there is nothing more to add.

Ailrun: Yeah, let's merge this first and then see whether we can remove that.

theories/Core/Semantic/PERLemmas.v
Ailrun: This one keeps a hypothesis that relates `A'` and `B`, which is not entirely identical. In those where two hypotheses are identical I clear one of them.

Issue #51, opened 2024-05-03
Get rid of indefinite description and propositional extensionality


HuStmpHrrr:
I think I just thought of a way to do it, but I actually need to put it down to see if it's actually the case.

PR #52, opened 2024-05-04
attempt to remove unnecessary axioms

HuStmpHrrr:
#51 

Please check my change carefully and see if it makes sense.

### DIFF
--- a/theories/Core/Semantic/PER.v
+++ b/theories/Core/Semantic/PER.v
@@ -15,12 +15,6 @@ Arguments mk_rel_mod_eval {_ _ _ _ _ _}.
 Inductive rel_mod_app (R : relation domain) f a f' a' : Prop := mk_rel_mod_app : forall fa f'a', {{ $| f & a |↘ fa }} -> {{ $| f' & a' |↘ f'a' }} -> {{ Dom fa ≈ f'a' ∈ R }} -> rel_mod_app R f a f' a'.
 Arguments mk_rel_mod_app {_ _ _ _ _}.
 
-Class IrrelevantParametrizedPER {A : Type} (IR : relation A) (OR : forall {a b}, IR a b -> relation A) : Prop := {
-    IPP_irrelevance : forall {a b} (ab ab' : IR a b), OR ab = OR ab';
-    IPP_sym : forall {a b} (ab : IR a b) (ba : IR b a), forall c d, OR ab c d -> OR ba d c;
-    IPP_trans : forall {a1 a2 a3} (a12 : IR a1 a2) (a23 : IR a2 a3) (a13 : IR a1 a3), forall b1 b2 b3, OR a12 b1 b2 -> OR a23 b2 b3 -> OR a13 b1 b3;
-  }.
-
 Definition per_bot : relation domain_ne := fun m n => (forall s, exists L, {{ Rne m in s ↘ L }} /\ {{ Rne n in s ↘ L }}).
 Arguments per_bot /.
 
@@ -60,7 +54,6 @@ Section Per_univ_elem_core_def.
          (out_rel : forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), relation domain)
          (equiv_a_a' : {{ DF a ≈ a' ∈ per_univ_elem_core ↘ in_rel}}),
           PER in_rel ->
-          IrrelevantParametrizedPER in_rel (@out_rel) ->
           (forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}),
               rel_mod_eval per_univ_elem_core B d{{{ p ↦ c }}} B' d{{{ p' ↦ c' }}} (out_rel equiv_c_c')) ->
           (forall f f', elem_rel f f' = forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), rel_mod_app (out_rel equiv_c_c') f c f' c') ->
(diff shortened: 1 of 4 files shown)


theories/Core/Semantic/PERLemmas.v
```diff
@@ -183,15 +184,15 @@ Lemma per_univ_elem_left_irrel : forall i A B R A' R',
...
-  specialize (H2 a b).
-  apply propositional_extensionality; firstorder.
+  intros.
+  apply per_univ_elem_sym in H.
+  apply per_univ_elem_sym in H0.
+  destruct_all.
+  eauto using per_univ_elem_right_irrel.
 Qed.
 
+(* JH: the code below has not been fixed yet *)
```
HuStmpHrrr, line 194: fixes are only until this point. let's see if it makes sense first. the idea is to impose structure.

theories/Core/Semantic/PER.v
```diff
@@ -15,6 +15,12 @@ Arguments mk_rel_mod_eval {_ _ _ _ _ _}.
 Inductive rel_mod_app (R : relation domain) f a f' a' : Prop := mk_rel_mod_app : forall fa f'a', {{ $| f & a |↘ fa }} -> {{ $| f' & a' |↘ f'a' }} -> {{ Dom fa ≈ f'a' ∈ R }} -> rel_mod_app R f a f' a'.
 Arguments mk_rel_mod_app {_ _ _ _ _}.
 
+Class IrrelevantParametrizedPER {A : Type} (IR : relation A) (OR : forall {a b}, IR a b -> relation A) : Prop := {
+    IPP_irrelevance : forall {a b} (ab ab' : IR a b), OR ab = OR ab';
```
HuStmpHrrr, line 19: We could relax this to an equivalence relation.

theories/Core/Semantic/PER.v
```diff
@@ -138,6 +148,8 @@ Section Per_univ_elem_ind_def.
         (out_rel : forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), relation domain),
         {{ DF A ≈ A' ∈ per_univ_elem i ↘ in_rel }} ->
         motive i A A' in_rel ->
+        PER in_rel ->
+        IrrelevantParametrizedPER in_rel (@out_rel) ->
         (forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}),
             rel_mod_eval (fun x y R => {{ DF x ≈ y ∈ per_univ_elem i ↘ R }} /\ motive i x y R) B d{{{ p ↦ c }}} B' d{{{ p' ↦ c' }}} (out_rel equiv_c_c')) ->
         (forall f f', elem_rel f f' = forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), rel_mod_app (out_rel equiv_c_c') f c f' c') ->
```
HuStmpHrrr, line 155: I am also a bit concern about the equality here. should it be just equivalence?

HuStmpHrrr:
after thinking a bit, I think we don't even need functional extensionality, if we always use equivalence between `Prop`.

theories/Core/Semantic/PER.v
```diff
@@ -138,6 +140,7 @@ Section Per_univ_elem_ind_def.
         (out_rel : forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), relation domain),
         {{ DF A ≈ A' ∈ per_univ_elem i ↘ in_rel }} ->
         motive i A A' in_rel ->
+        PER in_rel ->
```
Ailrun, line 143: Once `per_univ_elem_trans` on this extended definition is proved we can basically remove this argument, right? Like, by using `{{ DF A ≈ A' ∈ per_univ_elem i ↘ in_rel }}` and the symmetry&transitivity lemmas.

theories/Core/Semantic/PERLemmas.v
HuStmpHrrr: done

theories/Core/Semantic/PER.v
HuStmpHrrr: removed

theories/Core/Semantic/PER.v
HuStmpHrrr: yes, this is the same as strengthening IH. we can prove a new constructor which constructs PER from symmetry and transitivity.

HuStmpHrrr:
@Ailrun if you feel good, please go ahead and merge it. the next step is cumulativity and the corresponding lemmas for contexts and substitutions (with potential adjustments). The latter part can go in parallel with the realizability proof. I think @Antoine-something can also help get things done in par. maybe you can work out the lemmas for contexts and substitutions? realizability is a bit more difficult so it can be left for Clare.

theories/Core/Semantic/PERLemmas.v
```diff
@@ -183,95 +191,103 @@ Lemma per_univ_elem_left_irrel : forall i A B R A' R',
...
+Lemma per_univ_elem_cross_irrel2 : forall i A B R A' R',
+    per_univ_elem i A B R ->
+    per_univ_elem i B A' R' ->
+    R = R'.
+Proof.
+  intros.
+  apply per_univ_elem_sym in H.
+  destruct_all.
+  eauto using per_univ_elem_right_irrel.
+Qed.
```
Ailrun, line 221: This lemma is basically with same arguments but in the reversed order to the previous lemma. We can remove this.

HuStmpHrrr: yeah you are right.

HuStmpHrrr: removed

theories/Core/Semantic/PERLemmas.v
```diff
@@ -183,95 +191,103 @@ Lemma per_univ_elem_left_irrel : forall i A B R A' R',
...
+    | H1 : {{ DF ~?A ≈ ~_ ∈ per_univ_elem ?i ↘ ?R1 }},
+        H2 : {{ DF ~_ ≈ ~?A ∈ per_univ_elem ?i ↘ ?R2 }} |- _ =>
+        tryif unify R1 R2 then fail else
+          (let H := fresh "H" in assert (R1 = R2) as H by (eapply per_univ_elem_cross_irrel1; eassumption); subst;
+                                 try rewrite <- H in *)
+    | H1 : {{ DF ~_ ≈ ~?B ∈ per_univ_elem ?i ↘ ?R1 }},
+        H2 : {{ DF ~?B ≈ ~_ ∈ per_univ_elem ?i ↘ ?R2 }} |- _ =>
+        tryif unify R1 R2 then fail else
+          (let H := fresh "H" in assert (R1 = R2) as H by (eapply per_univ_elem_cross_irrel2; eassumption); subst;
+                                 try rewrite <- H in *)
```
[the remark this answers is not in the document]
HuStmpHrrr: removed

theories/Core/Semantic/PER.v
HuStmpHrrr: no it shouldn't. it's fine like this. In fact, we can get rid of some use of extensionality by asking `elem_ref = fun f f' => ...`.

PR #60, opened 2024-05-06
Finish Realizability for PER

### LINKED ISSUE #35: Prove the realizability lemma for the PER model
c.f. https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Semantics/Realizability.agda

Ailrun:
Resolves #35

### DIFF
--- a/theories/Core/Semantic/Realize.v
+++ b/theories/Core/Semantic/Realize.v
@@ -1,0 +1,107 @@
+From Coq Require Import Lia PeanoNat Relation_Definitions.
+From Equations Require Import Equations.
+From Mcltt Require Import Base Domain Evaluate EvaluateLemmas LibTactics PER PERLemmas Syntax System.
+
+Lemma per_nat_then_per_top : forall {n m},
+    {{ Dom n ≈ m ∈ per_nat }} ->
+    {{ Dom ⇓ ℕ n ≈ ⇓ ℕ m ∈ per_top }}.
+Proof.
+  induction 1; simpl in *; intros.
+  - eexists; firstorder econstructor.
+  - specialize (IHper_nat s) as [? []].
+    eexists; firstorder (econstructor; eauto).
+  - specialize (H s) as [? []].
+    eexists; firstorder (econstructor; eauto).
+Qed.
+
+#[export]
+Hint Resolve per_nat_then_per_top : mcltt.
+
+Section Per_univ_elem_realize.
+  Variable (i : nat) (rec_per_univ_then_per_top_typ : forall j, j < i -> forall a a' R, {{ DF a ≈ a' ∈ per_univ_elem j ↘ R }} -> {{ Dom a ≈ a' ∈ per_top_typ }}).
+
+  Lemma realize_per_univ_elem_gen : forall a a' R,
+      {{ DF a ≈ a' ∈ per_univ_elem i ↘ R }} ->
+      {{ Dom a ≈ a' ∈ per_top_typ }}
+      /\ (forall {c c'}, {{ Dom c ≈ c' ∈ per_bot }} -> {{ Dom ⇑ a c ≈ ⇑ a' c' ∈ R }})
+      /\ (forall {b b'}, {{ Dom b ≈ b' ∈ R }} -> {{ Dom ⇓ a b ≈ ⇓ a' b' ∈ per_top }}).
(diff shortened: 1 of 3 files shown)


theories/Core/Semantic/Realize.v
```diff
@@ -0,0 +1,107 @@
...
+    eexists; firstorder (econstructor; eauto).
+  - specialize (H s) as [? []].
+    eexists; firstorder (econstructor; eauto).
+Qed.
+
+#[export]
+Hint Resolve per_nat_then_per_top : mcltt.
+
+Section Per_univ_elem_realize.
+  Variable (i : nat) (rec_per_univ_then_per_top_typ : forall j, j < i -> forall a a' R, {{ DF a ≈ a' ∈ per_univ_elem j ↘ R }} -> {{ Dom a ≈ a' ∈ per_top_typ }}).
```
HuStmpHrrr, line 21: why IH provided by the universe case is not enough? you shouldn't need this to prove realizability.

theories/_CoqProject
```diff
@@ -11,6 +11,7 @@
 ./Core/Semantic/PERLemmas.v
 ./Core/Semantic/Readback.v
 ./Core/Semantic/ReadbackLemmas.v
+./Core/Semantic/Realize.v
```
HuStmpHrrr, line 14: let's call this file "Realizability"

theories/Core/Semantic/Realize.v
HuStmpHrrr: ok I removed it. remember, the custom induction principle has handled the well-founded part.

theories/Core/Semantic/Realize.v
```diff
@@ -0,0 +1,104 @@
...
+#[export]
+Hint Resolve per_nat_then_per_top : mcltt.
+
+Section Per_univ_elem_realize.
+  Lemma realize_per_univ_elem_gen : forall i a a' R,
+      {{ DF a ≈ a' ∈ per_univ_elem i ↘ R }} ->
+      {{ Dom a ≈ a' ∈ per_top_typ }}
+      /\ (forall {c c'}, {{ Dom c ≈ c' ∈ per_bot }} -> {{ Dom ⇑ a c ≈ ⇑ a' c' ∈ R }})
+      /\ (forall {b b'}, {{ Dom b ≈ b' ∈ R }} -> {{ Dom ⇓ a b ≈ ⇓ a' b' ∈ per_top }}).
+  Proof with (solve [(((eexists; split) || idtac; econstructor) || idtac); mauto]).
```
HuStmpHrrr, line 26: why `idtac`? they are not useful here.

theories/_CoqProject
Ailrun: Other files are mostly in verb forms so that's why I named it like this.
Do you think it would be better to change them all in noun forms?

theories/Core/Semantic/Realize.v
Ailrun: Oh, this was no longer needed, true. It was a left-over thing from a previous proof, which is redacted because our function construct requires the type annotation for the argument.

theories/Core/Semantic/Realize.v
Ailrun: We can use `solve [mauto | econstructor; mauto | eexists; split; econstructor; mauto]`, but I don't want to repeat them.

theories/_CoqProject
HuStmpHrrr: let's change everything to noun. it's also like this in Agda. let's maintain a similar structure.

theories/Core/Semantic/Realize.v
Ailrun: or we can use `try`. I will use that.

theories/_CoqProject
Ailrun: Ok, let's do it in a separate PR.

HuStmpHrrr:
There should also be some clean up PRs. For example, `per_nat_then_per_top` should be placed elsewhere. If you feel that you are on fire, you could also swallow Antoine's cake if you want to @Ailrun .

Ailrun:
@HuStmpHrrr Why do you think so? They are all in the similar spirit, aren't they?

HuStmpHrrr:
at least this lemma and `var_per_bot` should live in the same file.

Ailrun:
`per_..._then_per_...` things are about the lattice structure of PERs given here, and `var_per_bot` is a different from them, at least from my point of view.

PR #67, opened 2024-05-08
Prove cumulativity

Ailrun:
Why is this so short...
I originally planed to add a separate module `Cumulativity.v` and expected something more involved, but guess what, this (less than) 20 line happened, and it does not make much sense to put just this single proofs (even with some corollaries) to live in a separate module.

### DIFF
--- a/theories/Core/Semantic/PERLemmas.v
+++ b/theories/Core/Semantic/PERLemmas.v
@@ -346,3 +346,19 @@ Proof with solve [mauto].
       firstorder (econstructor; eauto).
   - split; try per_univ_elem_econstructor...
 Qed.
+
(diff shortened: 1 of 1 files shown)


HuStmpHrrr:
> I am also a bit surprised, but it makes sense
> the theorem states that we have a fixed `R` to relate values, whereas in Agda, `R` is a result of recursion computed by `El`. In Agda, we cannot fix the relation between values but in Agda we can. I think it's a pleasant distinction caused by impredicativity.

HuStmpHrrr:
Closes #67

Ailrun:
Oops, I forgot to put that on the description 😅

PR #68, opened 2024-05-08
Prove PER for per_ctx_env

Ailrun:


theories/Core/Semantic/PERBasicLemmas.v
```diff
@@ -0,0 +1,354 @@
...
+
+Ltac per_univ_elem_irrel_rewrite :=
+  functional_eval_rewrite_clear;
+  do_per_univ_elem_irrel_rewrite;
+  clear_dups.
+
+Lemma per_univ_elem_trans : forall i A1 A2 R,
+    per_univ_elem i A1 A2 R ->
+    forall j A3,
+    per_univ_elem j A2 A3 R ->
```
Ailrun, line 309: This generalizes the previous statement, where universe level of all 3 was the same.

theories/Core/Semantic/PERBasicLemmas.v
```diff
@@ -0,0 +1,354 @@
...
+  inversion_clear Hnl.
+  econstructor...
+Qed.
+
+#[export]
+Hint Resolve per_ne_trans : mcltt.
+
+Lemma per_univ_elem_right_irrel : forall i i' A B R B' R',
+    per_univ_elem i A B R ->
+    per_univ_elem i' A B' R' ->
```
Ailrun, line 191: This has the same generalization.

theories/Core/Semantic/PERBasicLemmas.v
```diff
@@ -0,0 +1,354 @@
...
+Ltac per_univ_elem_right_irrel_rewrite1 :=
+  match goal with
+  | H1 : {{ DF ~?A ≈ ~?B ∈ per_univ_elem ?i ↘ ?R1 }}, H2 : {{ DF ~?A ≈ ~?B' ∈ per_univ_elem ?i ↘ ?R2 }} |- _ =>
+      clean replace R2 with R1 by (eauto using per_univ_elem_right_irrel)
+  end.
+Ltac per_univ_elem_right_irrel_rewrite := repeat per_univ_elem_right_irrel_rewrite1.  
+
+Lemma per_univ_elem_sym : forall i A B R,
+    per_univ_elem i A B R ->
+    per_univ_elem i B A R /\ (forall a b, {{ Dom a ≈ b ∈ R }} -> {{ Dom b ≈ a ∈ R }}).
```
Ailrun, line 224: I realized that with `PER in_rel` argument, we do not need to use that `<->` trick here.

theories/Core/Semantic/PERBasicLemmas.v
```diff
@@ -0,0 +1,354 @@
+From Coq Require Import Lia PeanoNat Relation_Definitions RelationClasses.
+From Mcltt Require Import Axioms Base Evaluation LibTactics PERDefinitions Readback.
+Import Domain_Notations.
+
+(* Lemma rel_mod_eval_ex_pull : *)
```
HuStmpHrrr, line 5: I don't think we no longer need these lemmas.

theories/Core/Semantic/PERBasicLemmas.v
HuStmpHrrr: I suppose the generalization is for free?

theories/Core/Semantic/PERBasicLemmas.v
HuStmpHrrr: for real? even in the function case?

theories/Core/Semantic/PERBasicLemmas.v
HuStmpHrrr: when do we need this generalization though? is it also for free?

HuStmpHrrr: why should these lemmas be considered basic? and why the lemmas for contexts are not?

theories/Core/Semantic/PERBasicLemmas.v
Ailrun: Ok, if you also think so, let's really remove them.

theories/Core/Semantic/PERBasicLemmas.v
Ailrun: Yeah, it works now (as CI shows)

theories/Core/Semantic/PERBasicLemmas.v
Ailrun: Like only with a tiny change in the proof, so basically free.

theories/Core/Semantic/PERBasicLemmas.v
HuStmpHrrr: yeah they are here for the existential claims which were removed.

theories/Core/Semantic/PERBasicLemmas.v
Ailrun: Actually they are all basic. I made the distinction for proofs that requires Realizability, it's just that I somehow mixed up the distinction while proving process. I will clear them out.

HuStmpHrrr: which one uses realizability?

HuStmpHrrr:
Let's follow a bit closer to Agda's module structure and bring some hierarchy here as well. Unlike OCaml, Coq knows folders so there is no reason to keep files in a flap structure and use longer filenames to distinguish what they are for.

theories/Core/Semantic/PERBasicLemmas.v
Ailrun: No one here yet. If I remember correctly, there were some in the Agda mechanization.

Ailrun:
Sure, like `PER` directory that contains `Lemmas` and `Definition`, and `PER` module itself reexport them?

HuStmpHrrr:
something along the line, as long as it makes the code easier to browse.

theories/Core/Semantic/PERBasicLemmas.v
HuStmpHrrr: only this https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Semantics/Properties/PER.agda?ref_type=heads#L589

if you want to you can kick this lemma elsewhere.

Ailrun: Oh well then it's better to give it a special name then give others "basic"

Ailrun:
Let's do that refactoring in a different PR.

PR #75, opened 2024-05-10
what's the problem?

HuStmpHrrr:
What exactly is going wrong here?

### DIFF
--- a/theories/Core/Semantic/PER/Lemmas.v
+++ b/theories/Core/Semantic/PER/Lemmas.v
@@ -333,6 +333,38 @@ Proof.
   eassumption.
 Qed.
 
+Require Import Coq.Classes.RelationClasses.
+
+#[global]
+  Instance per_elem_PER {i A B R} `(H : per_univ_elem i A B R) : PER R | 10.
+Proof.
+  split.
+  - auto using (per_elem_sym _ _ _ _ _ _ H).
+  - eauto using (per_elem_trans _ _ _ _ _ _ _ H).
+Qed.
+
+
+#[global]
+  Instance per_univ_PER {i R} : PER (fun A B => per_univ_elem i A B R).
+Proof.
+  split.
+  - auto using per_univ_sym.
+  - eauto using per_univ_trans.
+Qed.
+
+
+Goal forall i A B R a b, per_univ_elem i A B R -> R a b.
+Proof.
+  intros.
+  assert (PER R). {
+    Print Instances PER.
+    Typeclasses eauto := debug.
+    Set Typeclasses Debug.
+    typeclasses eauto.
+  }
+  etransitivity.
+Abort.
+
 Lemma per_univ_elem_cumu : forall {i a0 a1 R},
     {{ DF a0 ≈ a1 ∈ per_univ_elem i ↘ R }} ->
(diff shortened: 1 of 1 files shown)


theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -333,6 +333,38 @@ Proof.
...
+
+
+Goal forall i A B R a b, per_univ_elem i A B R -> R a b.
+Proof.
+  intros.
+  assert (PER R). {
+    Print Instances PER.
+    Typeclasses eauto := debug.
+    Set Typeclasses Debug.
+    typeclasses eauto.
```
HuStmpHrrr, line 363: Type class resolution fails.

HuStmpHrrr: `eauto with typeclass_instances` succeeds. anyway to fix this? typeclass resolution doesn't look at the context at all?

Ailrun: You can check the trace with `typeclasses eauto || idtac`. For this specific case, the following is the trace:
```
Debug: 1: looking for (PER R) without backtracking
Debug: 1.1: simple eapply @per_elem_PER on (PER R), 1 subgoal(s)
Debug: 1.1-1 : (per_univ_elem ?i ?A ?B R)
Debug: 1.1-1: looking for (per_univ_elem ?i ?A ?B R) with backtracking
Debug: 1.1-1: no match for (per_univ_elem ?i ?A ?B R), 0 possibilities
Debug: 1.2: simple apply @Equivalence_PER on (PER R), 1 subgoal(s)
Debug: 1.2-1 : (Equivalence R)
Debug: 1.2-1: looking for (Equivalence R) without backtracking
Debug: 1.2-1: no match for (Equivalence R), 0 possibilities
```
In other words, it fails to find `per_univ_elem ?i ?A ?B R)`, unlike `eauto`. So indeed, it seems like it ignores the local context.

Ailrun: But the documentation says like this...
> When considering local hypotheses, we use the transparent state of the first hint database given. Using an empty database (created with [Create HintDb](https://coq.inria.fr/doc/V8.17.1/refman/proofs/automatic-tactics/auto.html#coq:cmd.Create-HintDb) for example) with unfoldable variables and [constants](https://coq.inria.fr/doc/V8.17.1/refman/language/core/definitions.html#term-constant) as the first argument of typeclasses eauto hence makes resolution with the local hypotheses use full conversion during unification.
which is confusing.

HuStmpHrrr: yeah I don't get it. It might just look for classes in the local contexts.

HuStmpHrrr: I want to add 

```

#[global]
  Hint Extern 1 => eassumption : typeclass_instances.
```

What do you think?

Ailrun: Sounds OK, we don't need to use `typeclasses eauto` in the proof very often anyway.

HuStmpHrrr: It's also very tempting to swap the order of `per_univ_elem`. How many times do we have to use PER properties later?

PR #79, opened 2024-05-14
define type class instances for PERs

### LINKED ISSUE #78: Define instances for the type class PER for the PER model

HuStmpHrrr:
Closes #78

### DIFF
--- a/theories/Core/Semantic/PER/Lemmas.v
+++ b/theories/Core/Semantic/PER/Lemmas.v
@@ -1,4 +1,5 @@
 From Coq Require Import Lia PeanoNat Relation_Definitions RelationClasses.
+From Equations Require Import Equations.
 From Mcltt Require Import Axioms Base Evaluation LibTactics PER.Definitions Readback.
 Import Domain_Notations.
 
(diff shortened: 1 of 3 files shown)


theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -499,3 +535,21 @@ Proof.
...
+#[global]
+  Instance per_ctx_PER {R} : PER (per_ctx_env R).
+Proof.
+  split.
+  - auto using per_ctx_sym.
+  - eauto using per_ctx_trans.
+Qed.
+
+
+#[global]
```
Ailrun, line 549: Why do you use `global` instead of `export` here?
(FYI, `global` was default up to Coq 8.17 and `export` is now after 8.18)

Ailrun:
Other than the attribute, LGTM.

theories/Core/Semantic/PER/Lemmas.v
HuStmpHrrr: Should be export I agree

Ailrun: It seems like Instance attributes are still `global`?

HuStmpHrrr: done

PR #81, opened 2024-05-15
Change isomorphism between per_elem

Ailrun:
Surprisingly, `per_univ_elem_cumu` is still short (almost does not change at all in fact)


theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -149,130 +149,178 @@ Qed.
 #[export]
 Hint Resolve per_ne_trans : mcltt.
 
-Lemma per_univ_elem_right_irrel : forall i i' R A B R' B',
-    per_univ_elem i R A B ->
-    per_univ_elem i' R' A B' ->
-    R = R'.
+Lemma per_univ_elem_respect_iff : forall i R a a' R',
+    {{ DF a ≈ a' ∈ per_univ_elem i ↘ R }} ->
+    (forall m m', {{ Dom m ≈ m' ∈ R }} <-> {{ Dom m ≈ m' ∈ R' }}) ->
```
HuStmpHrrr, line 154: I would use a notation to represent equivalence between relations, e.g. `R <==> R'`, or something matches your taste.

HuStmpHrrr:
use this PR to keep track of blockers, if any. (I am hoping not, as thinking about it right now, it is actually quite natural)

theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -149,130 +149,178 @@ Qed.
 #[export]
 Hint Resolve per_ne_trans : mcltt.
 
-Lemma per_univ_elem_right_irrel : forall i i' R A B R' B',
-    per_univ_elem i R A B ->
-    per_univ_elem i' R' A B' ->
-    R = R'.
+Lemma per_univ_elem_respect_iff : forall i R a a' R',
```
HuStmpHrrr, line 152: this lemma is useful for a setoid rewrite. we can proof an instance so that it will work with `setoid_rewrite` later.

theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -149,130 +149,178 @@ Qed.
...
+    intros;
+    intuition.
+  destruct_rel_mod_eval.
+  econstructor...
+Qed.
+
+#[export]
+Hint Resolve per_univ_elem_respect_iff : mcltt.
+
+Lemma per_univ_elem_right_irrel : forall i i' R a b R' b',
```
HuStmpHrrr, line 169: this was the only lemma using extensionality. it seems we don't need to use extensionality at all, at least for now. consider removing the Require above as well.

theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -149,130 +149,178 @@ Qed.
...
-        (* Order matters less here as H1 and H2 cannot be exchanged *)
-        clean replace R2 with R1 by first [solve [symmetry; eauto using per_univ_elem_cross_irrel] | tactic_error R2 R1]
-    end.
+  | H1 : {{ DF ~?A ≈ ~_ ∈ per_univ_elem ?i ↘ ?R1 }},
+      H2 : {{ DF ~?A ≈ ~_ ∈ per_univ_elem ?i' ↘ ?R2 }} |- _ =>
+      assert_fails (unify R1 R2);
+      assert (forall m m', R1 m m' <-> R2 m m') by (eapply per_univ_elem_right_irrel; [apply H1 | apply H2]) || tactic_error R1 R2;
+      let H := fresh "H" in
+      assert (forall m m', R2 m m' <-> R1 m m') by (eapply per_univ_elem_right_irrel; [apply H2 | apply H1]) || tactic_error R2 R1;
+      fail_if_dup;
```
HuStmpHrrr, line 297: would it not be helpful to apply the respectful lemma right now? then the proof context only needs to work with one relation. also, why prove two symmetric assertions?

theories/Core/Semantic/PER/Lemmas.v
Ailrun: Yeah we can provide morphism for `<->`

theories/Core/Semantic/PER/Lemmas.v
Ailrun: I wrote like that once, but realized that there are some cases that makes proofs more tedious. So I changed it like this.

HuStmpHrrr: how's that? we used to just do replace before. it should just be the same spirit?

theories/Core/Semantic/PER/Lemmas.v
Ailrun: Yeah that would be easier to read/understand

theories/Core/Semantic/PER/Lemmas.v
Ailrun: Before, it was basically `subst`, so all occurences change all together. Now, it's mere `<->`, so there are some "left-over" hypotheses that still use `R2`.

Ailrun: If the goal requires `H2` and one of such left-over hypotheses, it's easier to have matching relation then applying `..._respect` again.

HuStmpHrrr: so do we always apply the respectful lemma above? sounds cumbersome though. if you think it should be like this, then at last you can just use a lazymatch to find a symmetrized assumption, instead of proving it again and fail it.

HuStmpHrrr: something like `lazymatch goal with | R2 <==> R1 |- _ => fail 1 | |- _ => idtac end` should work.

Ailrun: True, that would work better. And if you think it would be very common to apply the respectful lemma on all instances of the pairs, I can add an option like that to this tactic.
However, as you can see in the following proofs, the respectful lemma is required only for specific pair (like the one required for the goal), so I'm not completely sure whether that option would be really helpful or not.

theories/Core/Semantic/PER/Definitions.v
```diff
@@ -8,6 +8,8 @@ Notation "'Dom' a ≈ b ∈ R" := ((R a b : Prop) : Prop) (in custom judg at lev
 Notation "'DF' a ≈ b ∈ R ↘ R'" := ((R R' a b : Prop) : Prop) (in custom judg at level 90, a custom domain, b custom domain, R constr, R' constr).
 Notation "'Exp' a ≈ b ∈ R" := (R a b : (Prop : Type)) (in custom judg at level 90, a custom exp, b custom exp, R constr).
 Notation "'EF' a ≈ b ∈ R ↘ R'" := (R R' a b : (Prop : Type)) (in custom judg at level 90, a custom exp, b custom exp, R constr, R' constr).
+Notation "R ~> R'" := (forall x y, R x y -> R' x y) (at level 90).
+Notation "R <~> R'" := (forall x y, R x y <-> R' x y) (at level 90).
```
HuStmpHrrr, line 12: is `<~>` equivalent to a conjunction of two `~>`?

theories/LibTactics.v
```diff
@@ -159,3 +159,6 @@ Ltac mautosolve := unshelve solve [mauto]; solve [constructor].
 
 #[export]
   Hint Extern 1 => eassumption : typeclass_instances.
+
+(* intuition tactic default setting *)
+Ltac Tauto.intuition_solver ::= auto with mcltt core solve_subterm.
```
HuStmpHrrr, line 164: What's the purpose of this line? It does way more than intuitionistic solving.

theories/Core/Semantic/Realizability.v
```diff
@@ -24,17 +24,22 @@ Lemma realize_per_univ_elem_gen : forall {i a a' R},
...
-  - eexists.
+  - subst; repeat econstructor.
+  - subst.
+    (on_all_hyp: fun H => rewrite -> H).
+    eexists.
     per_univ_elem_econstructor...
-  - destruct_by_head per_univ.
-    specialize (H1 _ _ _ H2).
+  - subst.
+    match_by_head1 elem_rel ltac:(fun H => on_all_hyp: fun H' => setoid_rewrite H' in H).
```
HuStmpHrrr, line 33: let's think of a better way to fix code like this. I am very hesitant to see lambda in Ltac used within the proof script.

theories/LibTactics.v
Ailrun: Original value for that is `auto with *`. This one is a restriction for that.

theories/Core/Semantic/Realizability.v
Ailrun: If we want a generic abstraction, higher order tactic is unavoidable IMO.

Ailrun: We can provide a specific ltac for each case, but that would still use the abstraction in the definition

theories/Core/Semantic/PER/Definitions.v
Ailrun: Logically yes, although they are not identical.

theories/Core/Semantic/PER/Lemmas.v
Ailrun: Done

theories/Core/Semantic/Realizability.v
Ailrun: I defined a specialized tactic (although its implementation still uses the HO tactic)

Ailrun:
Now proofs are fully converted

theories/Core/Completeness/FunctionCases.v
```diff
@@ -0,0 +1,69 @@
...
+From Mcltt Require Import Base LibTactics LogicalRelation System.
+Import Domain_Notations.
+
+Lemma rel_exp_pi_cong : forall {i Γ A A' B B'},
+  {{ Γ ⊨ A ≈ A' : Type@i }} ->
+  {{ Γ , A ⊨ B ≈ B' : Type@i }} ->
+  {{ Γ ⊨ Π A B ≈ Π A' B' : Type@i }}.
+Proof.
+  pose proof (@relation_equivalence_pointwise domain).
+  pose proof (@relation_equivalence_pointwise env).
```
Ailrun, line 11: Do you have any idea why `setoid_rewrite` for some instances work only with these explicit local hypotheses?

HuStmpHrrr:
I see that it's no longer WIP. let's review this PR and merge it first before adding more cases.

theories/Core/Semantic/PER/Definitions.v
```diff
@@ -8,42 +8,25 @@ Notation "'Dom' a ≈ b ∈ R" := ((R a b : Prop) : Prop) (in custom judg at lev
 Notation "'DF' a ≈ b ∈ R ↘ R'" := ((R R' a b : Prop) : Prop) (in custom judg at level 90, a custom domain, b custom domain, R constr, R' constr).
 Notation "'Exp' a ≈ b ∈ R" := (R a b : (Prop : Type)) (in custom judg at level 90, a custom exp, b custom exp, R constr).
 Notation "'EF' a ≈ b ∈ R ↘ R'" := (R R' a b : (Prop : Type)) (in custom judg at level 90, a custom exp, b custom exp, R constr, R' constr).
+(* Precedences of the next notations follow the ones in the standard library.
+   However, we do not use the ones in the standard library so that we can change
+   the relation if necessary in the future. *)
+Notation "R ~> R'" := (subrelation R R') (at level 70, right associativity).
+Notation "R <~> R'" := (relation_equivalence R R') (at level 95, no associativity).
```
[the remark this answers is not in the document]
Ailrun: Basically same, but some morphisms are pre-defined.

theories/Core/Semantic/Realizability.v
```diff
@@ -22,49 +22,49 @@ Lemma realize_per_univ_elem_gen : forall {i a a' R},
     /\ (forall {c c'}, {{ Dom c ≈ c' ∈ per_bot }} -> {{ Dom ⇑ a c ≈ ⇑ a' c' ∈ R }})
     /\ (forall {b b'}, {{ Dom b ≈ b' ∈ R }} -> {{ Dom ⇓ a b ≈ ⇓ a' b' ∈ per_top }}).
 Proof with (solve [try (try (eexists; split); econstructor); mauto]).
-  intros * H. simpl in H.
-  induction H using per_univ_elem_ind; repeat split; intros.
-  - subst...
-  - eexists.
+  pose proof (@relation_equivalence_pointwise domain).
```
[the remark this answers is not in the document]
Ailrun: I tried, and it didn't unfortunately.

theories/Core/Completeness/FunctionCases.v
Ailrun: I found this solution from stackoverflow,
https://stackoverflow.com/questions/73585260/setoid-rewrite-failed-with-mathclasses-coq
and I guess it contains (although it depends on some library) somewhat small repro?

HuStmpHrrr: Yes. The solution is not very satisfactory I guess. It's annoying that it doesn't work, though I suppose there aren't much more use later in the proof?

Ailrun: IDK, it is quite often required. Maybe I can hide them inside of the rewrite-based tactics but still...

HuStmpHrrr: it's ok. it is what it is.

PR #83, opened 2024-05-18
make CI parallel

### LINKED ISSUE #82: Parallelizing CI job
The current CI is single threaded. 

https://github.com/Beluga-lang/McLTT/blob/023acf26fc65642b7b424ede0c1aeefe73e3d146/.github/workflows/ci_build.yml#L55

we should make this `make` run with multiple threads.

https://github.com/SimenB/github-actions-cpu-cores

@Ailrun let me know if you have idea how to use this workflow.

HuStmpHrrr:
Closes #82

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -26,6 +26,9 @@ jobs:
     name: Continuous Intergration
     runs-on: ubuntu-latest
     steps:
+      - name: Get number of CPU cores
+        uses: SimenB/github-actions-cpu-cores@v2
+        id: cpu-cores
       - name: repo checkout
         uses: actions/checkout@v3
       - name: setup OCaml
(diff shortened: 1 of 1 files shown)


Ailrun:
Let's see how much it boosts the CI speed

HuStmpHrrr:
A few minutes. I suppose the bottleneck isn't the proof checking yet. Installing OCaml and then Coq and then plugins is very slow. Is there any way we can cache this setup?

HuStmpHrrr:
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

do we know where the folders are?

HuStmpHrrr:
are we able to use this action instead? https://github.com/coq-community/docker-coq-action

Ailrun:
I'm not sure that the first one would speed things up that much. Ocaml-setup time would  remain similar for the first (with good `~/.opam/<switch>` cache), and GH Actions cache is not so fast (it takes about a min) so the save would be about a min. The second approach might work in a higher chance, I guess (or in the worst case we can use the coq docker image directly)

PR #84, opened 2024-05-18
More completeness cases

Ailrun:


theories/Core/Completeness/FunctionCases.v
```diff
@@ -36,7 +44,7 @@ Proof.
   exists (per_univ i).
   split; [> econstructor; only 1-2: repeat econstructor; eauto ..].
   unfold per_univ.
-  eexists.
+  eexists ?[elem_rel].
   per_univ_elem_econstructor; try (now eauto); try setoid_reflexivity.
```
HuStmpHrrr, line 48: consider using the helpful lemma instead. then you can instantiate `elem_rel` when you apply the lemma, so you don't have to use `instantiate` below.

Ailrun: Oh this `elem_rel` is a workaround for the `reflexivity` later. But even for `out_rel`, that's a good suggestion.

theories/Core/Completeness/FunctionCases.v
```diff
@@ -60,13 +64,10 @@ Proof.
...
-    match goal with
-    | |- per_univ_elem _ ?R _ _ => setoid_replace R with elem_rel; subst elem_rel; [eassumption |]
-    end.
+    apply -> per_univ_elem_morphism_iff; eauto.
     split; intros; handle_per_univ_elem_irrel; intuition.
   - match goal with
-    | |- ?X <~> ?Y => instantiate (1 := Y)
+    | |- ?[elem_rel] <~> ?Y => instantiate (elem_rel := Y)
     end.
     reflexivity.
```
Ailrun, line 72: This part is where `elem_rel` is actually used

HuStmpHrrr: why do you need this match? doesn't reflexivity immediate solve it?

Ailrun: Unfortunately `reflexivity` goes into the reflexivity of `<->`, not of `<~>`, so it sees `?elem_rel f f' <-> [some complex predicate]`, so fails to find `?elem_rel` just by unification.
I can replace this with a lemma like `relation_equivalence_refl : ...` (or the similar pre-defined one inside of `relation_equivalence_equivalence`)

HuStmpHrrr: I am also certain that `reflexivity` works if you provide an instance for `Equivalence`.

Ailrun: No, the instance is pre-defined (but maybe it the same issue with `pose proof` things). It does not work with an instance.

HuStmpHrrr: it says it does: https://coq.inria.fr/doc/V8.17.1/refman/proofs/writing-proofs/equality.html#coq:tacn.symmetry

I no longer understand when Coq does what.

Ailrun: I mean yeah, it works sometimes, but when there are multiple possibility, I'm not sure what is the strategy they actually use. Like here, we have two choices:
1. `<~>` instance on `?elem_rel <~> fun f f' => [some complex predicate]`
2. `<->` instance on `?elem_rel f f' <-> [some complex predicate]`

Their strategy somehow prefers the 2.

HuStmpHrrr: weird. maybe set opaque for typeclass resolution?

Ailrun: It still doesn't work. I have no idea why really...

HuStmpHrrr: If so, there is nothing more I want to add. merge if you don't have other things to add to this PR.

HuStmpHrrr: I think I have a clue about what's happening. The unification during typeclass resolution is probably too syntactic. I will ask on discourse. This sucks.

PR #86, opened 2024-05-21
working on the logrel for soundness proof

HuStmpHrrr:


HuStmpHrrr:
The logical relations for soundness is somewhat cumbersome. In Agda code, they are defined by recursion on the PER model, which obviously is not possible in our case. I think it might just be easier to copy the PER model and extends it. Then we prove two lemmas: one says that a PER model necessarily gives a gluing model, and the other one says that a gluing model can be degenerated to a PER model. Hopefully, this is enough to recover the recursive definitions in Agda. I am not sure how repetitive the lemmas will be. I hope not by a lot after prove these two lemmas.

HuStmpHrrr:
It's ready for review. Let's prove properties in another PR.

theories/Core/Soundness/LogicalRelation.v
```diff
@@ -0,0 +1,343 @@
...
+  Definition univ_glu_pred' {j} (lt_j_i : j < i) : glu_pred :=
+    fun Γ m M A =>
+      {{ Γ ⊢ m : M }} /\ {{ Γ ⊢ M ≈ Type@j : Type@i }} /\
+        glu_univ_rec lt_j_i A A /\
+        glu_univ_typ_rec lt_j_i A Γ m.
+
+  #[global]
+    Arguments univ_glu_pred' {j} lt_j_i Γ m M A/.
+
+  Inductive glu_univ_elem_core : typ_pred -> glu_pred -> relation domain -> domain -> domain -> Prop :=
```
Ailrun, line 108: Should this return `relation domain`? What would be a problem if it contains `per_univ_elem` as an argument and then prove `glu_univ_elem ... -> per_univ ...`?

Ailrun: In particular, I think we anyway need to prove `glu_univ_elem ... elem_rel a b -> per_univ_elem ... elem_rel a b` if we do not want to add `PER in_rel` (as in `per_univ_elem` definition), and by removing `elem_rel` part out of `glu_univ_elem` we do not need to deal with it in the transitivity

HuStmpHrrr: > Should this return `relation domain`? What would be a problem if it contains `per_univ_elem` as an argument and then prove `glu_univ_elem ... -> per_univ ...`?

it needs to return a `typ_pred` and `glu_pred` because in Agda, these two are derived by recursion on the PER model. it makes no difference to take per_univ_elem as an argument as the premises will need to be listed again anyways.

HuStmpHrrr: > In particular, I think we anyway need to prove `glu_univ_elem ... elem_rel a b -> per_univ_elem ... elem_rel a b` if we do not want to add `PER in_rel` to `glu_univ_elem_core_pi` (as in `per_univ_elem_core` definition), and by removing `elem_rel` part out of `glu_univ_elem` we do not need to deal with it in the transitivity, among others.

I think we need to prove it anyways. I think it's not a major difficulty to reprove some lemmas. I think eventually this impredicative style will eventually combine both completeness and soundness, so ultimately there is only one gluing model for all. this work is necessary towards that direction.

Ailrun: > I think eventually this impredicative style will eventually combine both completeness and soundness

I'm not sure that would be a really good idea; as you can see in the `pr-finish-completeness` branch, natrec cases for completeness by themselves are already quite complex, so how much would the "combined" proof be readable/extensible.

Ailrun: so it would be good to have a way to factor the `per_univ_elem` part out really.

Ailrun: > it needs to return a typ_pred and glu_pred because in Agda

I mean the `elem_rel` part, not the other two parts, to be clear.

Ailrun: Specifically, I mean something like this:
```
  | glu_univ_elem_core_pi :
    `{ forall (in_rel : relation domain)
         IP IEl
         (out_rel : forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), relation domain)
         (OP : forall c c' (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), typ_pred)
         (OEl : forall c c' (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}), glu_pred)
         typ_rel el_rel
         (elem_rel : relation domain),
          per_univ_elem i in_rel a a' ->
          glu_univ_elem_core IP IEl a a' ->
          (forall {c c'} (equiv_c_c' : {{ Dom c ≈ c' ∈ in_rel }}),
              rel_mod_eval (fun R a b => glu_univ_elem_core (OP _ _ equiv_c_c') (OEl _ _ equiv_c_c') a b /\ per_univ_elem R a b)
                B d{{{ p ↦ c }}} B' d{{{ p' ↦ c' }}}
                (out_rel equiv_c_c')) ->
          per_univ_elem elem_rel d{{{ Π a p B }}} d{{{ Π a' p' B' }}} ->
          typ_rel <∙> pi_typ_pred i in_rel IP IEl OP ->
          el_rel <∙> pi_glu_pred i in_rel IP IEl elem_rel OEl ->
          glu_univ_elem_core typ_rel el_rel d{{{ Π a p B }}} d{{{ Π a' p' B' }}} }
```
(this still might have some duplicates due to `per_univ_elem elem_rel d{{{ Π a p B }}} d{{{ Π a' p' B' }}}`)
With this, we do not need to prove things about `elem_rel`. They are already done in `PERLemmas.`

HuStmpHrrr: done

Ailrun (approved):
LGTM

PR #88, opened 2024-05-21
More cases for completeness

### LINKED ISSUE #87: Finish Completeness
- [x] Cases related to universes
    - #74
    - #81
- [x] Cases related to structures of term typing
    - #74
    - #81
- [x] Cases related to functions
    - #81
    - #84
- [x] Cases related to variables
    - #74
    - #81
    - #88
- [x] Cases related to structures of substitutions
    - #88
- [x] Cases related to natural numbers
    - #88
    - #90
- [x] Fundamental theorem
    - #90
- [x] Main completeness theorem
    - #90

Ailrun:
Related: #87

Context, substitution, and variable cases are all done.
Now it's only natural number cases (which are most tedious and complex...)

However, I realized that the natural number cases requires a "stronger" functionality of `eval` (i.e. functionality modulo equivalence of environment), so I decided to make a separate PR after this.

### DIFF
--- a/theories/Core/Semantic/PER/Lemmas.v
+++ b/theories/Core/Semantic/PER/Lemmas.v
@@ -1,4 +1,4 @@
-From Coq Require Import Lia Morphisms Morphisms_Prop Morphisms_Relations PeanoNat Relation_Definitions RelationClasses.
+From Coq Require Import Equivalence Lia Morphisms Morphisms_Prop Morphisms_Relations PeanoNat Relation_Definitions RelationClasses.
 From Equations Require Import Equations.
 From Mcltt Require Import Axioms Base LibTactics.
 From Mcltt.Core Require Import Evaluation PER.Definitions PER.CoreTactics Readback.
@@ -184,7 +184,7 @@ Proof with mautosolve.
   simpl.
   intros R R' HRR'.
   split; intros Horig; [gen R' | gen R];
-    induction Horig using per_univ_elem_ind; per_univ_elem_econstructor; eauto;
+    induction Horig using per_univ_elem_ind; basic_per_univ_elem_econstructor; eauto;
     try (etransitivity; [symmetry + idtac|]; eassumption);
     intros;
     destruct_rel_mod_eval;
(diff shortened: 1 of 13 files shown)


theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -474,56 +474,105 @@ Proof.
...
+      | assert (in_rel c c') by intuition; (on_all_hyp: destruct_rel_by_assumption in_rel)];
+      econstructor; intuition.
+    destruct_by_head rel_typ.
+    handle_per_univ_elem_irrel.
+    intuition.
+Qed.
+
+Ltac invert_per_univ_elem H :=
+  (unshelve eapply (per_univ_elem_pi_clean_inversion _) in H; [| eassumption | |]; destruct H as [? []])
+  + basic_invert_per_univ_elem H.
```
Ailrun, line 536: I change the previous tactic like this so that it automatically uses `per_univ_elem_pi_clean_inversion`. Similar for `per_univ_elem_econstructor`.

Ailrun:
> However, I realized that the natural number cases requires a "stronger" functionality of eval (i.e. functionality modulo equivalence of environment), so I decided to make a separate PR after this.

Actually, this stronger functionality is not true (because of the `Π a p B` case which contains `p`). Maybe we need to bring functional extensionality again here to compare `p`s.

Ailrun:
Or we can define a weaker notion of equality between two domain values so that it compares that kind of environment occurence in a pointwise manner.

HuStmpHrrr:
where do we need to compare environments? I don't think we need that in Agda

Ailrun:
Now I think about it, maybe they are definitionally equal. The target was `drop_env (p \mapsto a \mapsto b)` and `p \mapsto a` if I remember correctly. (I hope it's not the other way around, i.e. dropping first and adding the top member)

HuStmpHrrr:
> Now I think about it, maybe they are definitionally equal. The target was `drop_env (p \mapsto a \mapsto b)` and `p \mapsto a` if I remember correctly. (I hope it's not the other way around, i.e. dropping first and adding the top member)

It should be this way. There is no essential need for functional extensionality in the Agda code.

Ailrun:
Yeah, anyway, let's pass this PR first and then finish those few natrec cases in a separate PR as this one is already quite big.

PR #97, opened 2024-06-03
Improve parallelism

Ailrun:


Ailrun:
Hm, this takes more, not less...

PR #99, opened 2024-06-04
need a few lemmas to push through this proof

HuStmpHrrr:


theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -73,19 +64,155 @@ Proof.
...
+    glu_univ_elem i P El A B ->
+    forall Γ t T a R,
+      El Γ t T a ->
+      per_univ_elem i R A B ->
+      R a a.
+Proof.
+  induction 1 using glu_univ_elem_ind; intros;
+    match goal with
+    | H : per_univ_elem _ _ _ _ |- _ => invert_per_univ_elem H
+    end;
```
Ailrun, line 203: You can use `match_by_head per_univ_elem invert_per_univ_elem`

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -73,19 +64,155 @@ Proof.
...
+  specialize (H16 a0 a0) as [? _].
+  specialize (H16 H7).
+  intros.
+  destruct_rel_mod_app.
+  destruct_rel_mod_eval.
+  functional_eval_rewrite_clear.
+  do_per_univ_elem_irrel_assert.
+
+  econstructor; eauto.
+  apply H23. trivial.
```
Ailrun, line 217: What do these empty lines mean? If they are to separate multiple cases, why aren't there `-`s (or similar goal focusing tools)?

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -73,19 +64,155 @@ Proof.
...
+      P Γ T ->
+      {{ ⊢ Γ ≈ Δ }} ->
+      P Δ T.
+Proof.
+  induction 1 using glu_univ_elem_ind; intros;
+    simpl_glu_rel; mauto 2.
+
+  - econstructor; mauto.
+
+  - mauto 6.
```
Ailrun, line 129: If `mauto 6` or `econstructor; mautosolve` works for both, how about `mauto2; <the working tactic>.`?

Ailrun (commented):
Some comments about proofs

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: for some reason, this loops. I think it's because for the Pi case, the inversion keeps creating new things to invert but in reality, I only need to do it twice.

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: It's clearer interactively. the previous tactics are saturation, and then solving.

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: done

theories/Core/Soundness/LogicalRelation/Lemmas.v
Ailrun: Oh then you want `match_by_head1` instead

theories/Core/Soundness/LogicalRelation/Lemmas.v
Ailrun: I see. In the solving part, would it be possible to avoid mentioning `H23` explicitly? That would make proof maintenance harder.

theories/Core/Syntactic/System/Definitions.v
```diff
@@ -334,3 +334,29 @@ Qed.
...
+    forall (σ σ' : sub) (M M' : typ),
+      {{ Γ ⊢s σ ≈ σ' : Δ }} ->
+      {{ Δ ⊢ A : Type @ i }} ->
+      {{ Γ ⊢ M ≈ M' : ~ A [ σ ] }} ->
+      {{ Γ ⊢s ~(e1 σ M) ≈ ~ (e2 σ' M') : Δ, A }}.
+Arguments a_extend_eq_cong_rel i Γ Δ A/.
+#[export]
+  Typeclasses Transparent a_extend_eq_cong_rel.
+
+(* this morphism is added this way because the intended ==> syntax does not allow to add the well-formedness of A *)
```
Ailrun, line 357: How about defining it as a parametric morphism?

theories/Core/Syntactic/System/Lemmas.v
```diff
@@ -153,33 +153,38 @@ Proof.
   mauto.
 Qed.
 
-Lemma exp_eq_sub_cong_typ1 : forall {Δ Γ A A' σ i}, {{ Δ ⊢ A ≈ A' : Type@i }} -> {{ Γ ⊢s σ : Δ }} -> {{ Γ ⊢ A[σ] ≈ A'[σ] : Type@i }}.
+Lemma exp_eq_sub_cong_typ1 : forall Δ Γ A A' σ i, {{ Δ ⊢ A ≈ A' : Type@i }} -> {{ Γ ⊢s σ : Δ }} -> {{ Γ ⊢ A[σ] ≈ A'[σ] : Type@i }}.
```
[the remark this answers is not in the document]
HuStmpHrrr: otherwise they cannot be added to a rewrite hint. I don't know why rewrite cares, but it is what it is.

theories/Core/Syntactic/System/Definitions.v
HuStmpHrrr: this is the problem, parametric morphism doesn't allow additional conditions.

theories/Core/Syntactic/System/Lemmas.v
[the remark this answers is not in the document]
Ailrun: Huh, that's weird...

theories/Core/Syntactic/System/Definitions.v
Ailrun: Reordering the argument and then defining the morphism as `a_extend TypingForA` would work, right? Hm. Maybe that's too tedious.

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -73,19 +66,212 @@ Proof.
...
+    assert {{ Δ ⊢ m' : IT [ σ ]}} by eauto using glu_univ_elem_trm_escape.
+    edestruct H11 as [? []]; eauto.
+    eexists; split; eauto.
+    eapply H2; eauto.
+    assert {{ Γ ⊢ m ≈ t' : Π IT OT }} as Hty by mauto.
+    assert {{ Δ ⊢ IT [ σ ] : Type @ i4 }} by mauto 3.
+    eapply wf_exp_eq_sub_cong with (Γ := Δ) in Hty; [|mauto 4].
+    autorewrite with mcltt in Hty.
+    eapply wf_exp_eq_app_cong with (N := m') (N' := m') in Hty; try pi_univ_level_tac; [|mauto 2].
+    autorewrite with mcltt in Hty.
```
HuStmpHrrr, line 272: @Ailrun I think these few lines are interesting to see. autorewrite drives through complex type equivalence reasoning and kills the goal.

theories/Core/Syntactic/Corollaries.v
```diff
@@ -35,3 +36,16 @@ Qed.
 
 #[export]
 Hint Resolve invert_sub_id : mcltt.
+
+Add Parametric Morphism i Γ Δ : a_sub
```
HuStmpHrrr, line 40: I have no idea why this rule doesn't work. I think it might have something to do with the failure to identify well-formed reflexive equivalence judgments. autorewrite in fact depends on the next morphism.

theories/Core/Syntactic/System/Definitions.v
HuStmpHrrr: yeah this morphism needs to be redefined. stay tuned.

theories/Core/Syntactic/System/Lemmas.v
```diff
@@ -42,6 +42,19 @@ Proof with mautosolve.
...
+  match goal with
+  | |- {{ ~_ ⊢s ~_ : ~_ }} => mauto 4
+  | H : {{ ~?Δ ⊢ ~?A : ~(a_typ ?j) }} |- {{ ~?Δ , ~?A ⊢ ~?B : ~(a_typ ?i) }} =>
+      eapply lift_exp_max_right; mauto 4
+  | |- {{ ~?Δ ⊢ ~?A : ~(a_typ ?j) }} =>
+      eapply lift_exp_max_left; mauto 4
+  end.
+
+#[export]
+  Hint Rewrite -> wf_exp_eq_pi_sub using pi_univ_level_tac : mcltt.
```
HuStmpHrrr, line 56: rewrites should always come with tactics that tries to solve their subgoals.

HuStmpHrrr:
let's try to merge what I have for now. more lemmas are coming in next PRs.

theories/Core/Syntactic/Corollaries.v
Ailrun: It cannot find reflexivity for this, as, afaik, generalized rewriting does not support PER very well.

Ailrun (approved):
Let's merge this. I think I can refactor some proofs later (so that there are less explicit hypothesis names).

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: done

PR #101, opened 2024-06-06
Add syntactic judgements for type subsumption

### LINKED ISSUE #100: Syntactic judgement that reflects cumulativity
Suppose the case of `rec M return idOfType4 Type@3 | zero -> Type@0 | succ -> Type@1`.
To infer `Type@3` or check against `Type@j` (where `j >= 3`), we need decidable checking that `idOfType4 Type@3` has some relation with `Type@1` (the minimum universe of `Type@0`) (and also with `Type@2` for `Type@1`).
This relation is not conversion; clearly their normal forms, `Type@3` and `Type@1`, are not convertible.

One way to deal with this is, after normalization (by evaluation), instead of checking syntactic equality between the normal forms of the inferred type and given type, we can check that the (normal form of) given type is obtained by applying cumulativity on the inferred type. Yet, we have a problem. To what is this check actually realized? We do not have a syntactic judgement that describes the notion of "conversion, but with directed cumulativity". In other words, we do not have a good return type for a function like
https://github.com/Beluga-lang/McLTT/blob/6f4fb0d35eb96fc3e6df92c97392d68744d93783/theories/Extraction/TypeCheck.v#L26-L35
because we do not have something like `{{ G ⊢ M ≈ M' : A }}` for "conversion, but with directed cumulativity".

Ailrun:
Resolves #100

### DIFF
--- a/theories/Core/Syntactic/System/Definitions.v
+++ b/theories/Core/Syntactic/System/Definitions.v
@@ -367,16 +367,13 @@ Proof.
   eauto using typ_subsumption_trans.
 Qed.
 
-Inductive nf_subsumption : nf -> nf -> Prop :=
+Variant nf_subsumption : nf -> nf -> Prop :=
 | nf_subsumption_typ :
-  `{ nf_subsumption n{{{ Type@i }}} n{{{ Type@(S i) }}} }
+  `{ i <= j ->
+     nf_subsumption n{{{ Type@i }}} n{{{ Type@j }}} }
 | nf_subsumption_eq :
   `{ A = A' ->
      nf_subsumption A A' }
-| nf_subsumption_trans :
-  `{ nf_subsumption A A' ->
-     nf_subsumption A' A'' ->
-     nf_subsumption A A'' }
 .
 
 #[export]
@@ -385,5 +382,8 @@ Hint Constructors nf_subsumption : mcltt.
 #[export]
 Instance nf_subsumption_Transitive : Transitive nf_subsumption.
 Proof.
-  eauto using nf_subsumption_trans.
+  intros A A' A'' [];
+    inversion 1; subst; mauto.
+  econstructor.
+  lia.
 Qed.
--- a/theories/Core/Syntactic/System/Definitions.v
+++ b/theories/Core/Syntactic/System/Definitions.v
@@ -26,6 +26,7 @@ Inductive wf_ctx : ctx -> Prop :=
      {{ Γ ⊢ A : Type@i }} ->
      {{ ⊢ Γ , A }} )
 where "⊢ Γ" := (wf_ctx Γ) (in custom judg) : type_scope
+
 with wf_ctx_eq : ctx -> ctx -> Prop :=
 | wf_ctx_eq_empty : {{ ⊢ ⋅ ≈ ⋅ }}
 | wf_ctx_eq_extend :
@@ -36,6 +37,7 @@ with wf_ctx_eq : ctx -> ctx -> Prop :=
      {{ Δ ⊢ A ≈ A' : Type@i }} ->
      {{ ⊢ Γ , A ≈ Δ , A' }} )
 where "⊢ Γ ≈ Γ'" := (wf_ctx_eq Γ Γ') (in custom judg) : type_scope
+
 with wf_exp : ctx -> exp -> typ -> Prop :=
 | wf_univ :
   `( {{ ⊢ Γ }} ->
@@ -85,6 +87,7 @@ with wf_exp : ctx -> exp -> typ -> Prop :=
   `( {{ Γ ⊢ A : Type@i }} ->
      {{ Γ ⊢ A : Type@(S i) }} )
 where "Γ ⊢ M : A" := (wf_exp Γ M A) (in custom judg) : type_scope
+
 with wf_sub : ctx -> sub -> ctx -> Prop :=
 | wf_sub_id :
   `( {{ ⊢ Γ }} ->
@@ -106,6 +109,7 @@ with wf_sub : ctx -> sub -> ctx -> Prop :=
      {{ ⊢ Δ ≈ Δ' }} ->
      {{ Γ ⊢s σ : Δ' }} )
 where "Γ ⊢s σ : Δ" := (wf_sub Γ σ Δ) (in custom judg) : type_scope
+
 with wf_exp_eq : ctx -> typ -> exp -> exp -> Prop :=
 | wf_exp_eq_typ_sub :
   `( {{ Γ ⊢s σ : Δ }} ->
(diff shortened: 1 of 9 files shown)


theories/Core/Syntactic/System/Definitions.v
```diff
@@ -333,9 +336,54 @@ Proof.
...
+#[export]
+Hint Constructors typ_subsumption : mcltt.
+
+#[export]
+Instance typ_subsumption_Transitive Γ : Transitive (typ_subsumption Γ).
+Proof.
+  eauto using typ_subsumption_trans.
+Qed.
+
+Inductive nf_subsumption : nf -> nf -> Prop :=
```
HuStmpHrrr, line 370: do we actually need this relation? and why?

theories/Core/Syntactic/System/Lemmas.v
```diff
@@ -673,3 +702,106 @@ Qed.
...
+  - induction H; mauto.
+    + right; do 2 eexists; firstorder.
+    + firstorder; subst; firstorder.
+      match_by_head @eq ltac:(fun H => inversion H; subst).
+      assert (x1 <= x0) by lia.
+      firstorder.
+  - firstorder; subst; mauto.
+Qed.
+
+Definition nf_subsumption_dec A A' :
```
HuStmpHrrr, line 793: will we need to extract this function?

theories/Core/Syntactic/System/Lemmas.v
```diff
@@ -673,3 +702,106 @@ Qed.
...
+  gen j i.
+  induction Hsub; intros; subst; mauto 4.
+  assert (exists j', {{ Γ ⊢ A' : Type@j' }}) as [] by mauto 2.
+  etransitivity; mautosolve 4.
+Qed.
+
+#[export]
+Hint Resolve wf_exp_and_nf_subsumption_implies_typ_subsumption : mcltt.
+
+Lemma nf_subsumption_spec : forall {A A'},
```
HuStmpHrrr, line 779: I think this is actually is a better definition than the definition above. There are only two cases needed.

theories/Extraction/TypeCheck.v
```diff
@@ -33,3 +33,27 @@ Equations conversion_dec G A M (HM : {{ G ⊢ M : A }}) M' (HM' : {{ G ⊢ M' :
     | right _ => right _
     end.
 Next Obligation. impl_obl_tac. Qed.
+
+Equations? typ_subsumption_dec G i A (HA : {{ G ⊢ A : Type@i }}) A' (HA' : {{ G ⊢ A' : Type@i }}) : { {{ G ⊢ A ⊆ A' }} } + { ~ {{ G ⊢ A ⊆ A' }} } :=
+| G, i, A, HA, A', HA' =>
+    let (W, HW) := nbe_impl G A {{{ Type@i }}} _ in
+    let (W', HW') := nbe_impl G A' {{{ Type@i }}} _ in
+    match nf_subsumption_dec W W' with
```
HuStmpHrrr, line 41: let's not use Definition to define this function if it's meant to be used.

theories/Core/Syntactic/System/Lemmas.v
Ailrun: Yes, it would be better to move this function to somewhere else.

theories/Core/Syntactic/System/Definitions.v
Ailrun: Otherwise, what should we return by comparing two `nf`s?

theories/Extraction/TypeCheck.v
Ailrun: You mean inlining the definition?

theories/Extraction/TypeCheck.v
```diff
@@ -33,3 +33,27 @@ Equations conversion_dec G A M (HM : {{ G ⊢ M : A }}) M' (HM' : {{ G ⊢ M' :
     | right _ => right _
     end.
 Next Obligation. impl_obl_tac. Qed.
+
+Equations? typ_subsumption_dec G i A (HA : {{ G ⊢ A : Type@i }}) A' (HA' : {{ G ⊢ A' : Type@i }}) : { {{ G ⊢ A ⊆ A' }} } + { ~ {{ G ⊢ A ⊆ A' }} } :=
```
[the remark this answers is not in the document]
Ailrun: We can add a wrapper around this like ` typ_subsumption_dec' G A i (HA : {{ G ⊢ A : Type@i }}) A' i' (HA' : {{ G ⊢ A' : Type@i' }} : ...` and in the definition call `typ_subsumption_dec` with `max i i'`, so yeah, it is possible, but also can be done after defining this.

theories/Core/Syntactic/System/Lemmas.v
Ailrun: I thought that this one might be more cumbersome for `wf_exp_and_nf_subsumption_implies_typ_subsumption`, but it seems not after actually proving that. I will change the definition.

theories/Extraction/TypeCheck.v
Ailrun: Anyway, I replaced one with `Equations`.

Ailrun:
Everything I'd like to put in this PR is done.

theories/Extraction/TypeCheck.v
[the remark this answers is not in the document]
HuStmpHrrr: remember to remove the question mark as well. this could generate more obligations.

theories/Extraction/TypeCheck.v
HuStmpHrrr: You don't see the body directly in Coq. It also gives troubles if you ever need to prove anything about it.

theories/Core/Syntactic/System/Lemmas.v
HuStmpHrrr: this definition is definitively simpler because you don't have to handle general transitivity. I think the type subsumption relation can also be improved. two types should either be convertible, or cumulated if they are both universes. In this way, transitivity is also removed.

theories/Extraction/TypeCheck.v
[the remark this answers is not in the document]
Ailrun: Whatkind of more obligations? I haven't observed any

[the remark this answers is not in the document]
HuStmpHrrr: This particular definition is simple enough but in general the question mark is only for development. It's not meant to stay there.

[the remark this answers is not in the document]
Ailrun: Is it? I mean, the other MLTT mechanization uses `?` mainly, and that style seems better to me.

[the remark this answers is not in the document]
HuStmpHrrr: Question marks allow you to use proof mode to figure out how you want to fill in the holes. No equations or induction principles are generated, for example.

PR #107, opened 2024-06-09
Add helper lemmas

### LINKED ISSUE #32: helper rules
In my agda proof, https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Statics/Full.agda, there are rules that are marked as `remove after presupposition`. Proves these rules as lemmas. They can be added to the hint database.

Ailrun:
Resolve #32 and refactor some imports

### DIFF
--- a/theories/Core/Syntactic/SystemOpt.v
+++ b/theories/Core/Syntactic/SystemOpt.v
@@ -1,0 +1,234 @@
+From Coq Require Import Setoid.
+From Mcltt Require Import Base LibTactics.
+From Mcltt.Core Require Export CtxEq Presup System.
+Import Syntax_Notations.
+
+#[local]
+Ltac impl_opt_constructor :=
+  intros;
+  gen_presups;
+  mautosolve 4.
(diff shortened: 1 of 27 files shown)


theories/Core/Syntactic/SystemOpt.v
```diff
@@ -0,0 +1,234 @@
...
+  impl_opt_constructor.
+Qed.
+
+#[export]
+Hint Resolve wf_fn' : mcltt.
+#[export]
+Remove Hints wf_fn : mcltt.
+
+Corollary wf_app' : forall {Γ M N A B i},
+    {{ Γ , A ⊢ B : Type@i }} ->
```
HuStmpHrrr, line 52: I don't think we need this premise at all. we can prove a helper lemma to show a well-formed Pi must imply a well-formed B.

Ailrun: Ah right. Let me remove that premise.

PR #122, opened 2024-06-20
Working on subtyping

### LINKED ISSUE #121: Add subtyping for cumulativity
C.f. #120

Let's implement subtyping

- [x] syntactic extension
    - #122
    - #127
- [x] PER model
    - ~~#125~~
    - #128
- [x] PER model lemma
    - #128
- [x] completeness
    - #130 
    - #133
- [x] soundness
    - #140 
    - #153
- [x] subtyping checker
    - #124
    - #152

HuStmpHrrr:
C.f. #121 

This PR is likely to be huge. Stay tuned.

### DIFF
--- a/theories/Core/Completeness/Consequences/Types.v
+++ b/theories/Core/Completeness/Consequences/Types.v
@@ -87,148 +87,149 @@ Hint Resolve is_typ_constr_and_exp_eq_nat_implies_eq_nat : mcltt.
 
 (* We cannot use this spec as the definition of [typ_subsumption] as
    then its transitivity requires [exp_eq_typ_implies_eq_level] or a similar semantic lemma *)
-Lemma typ_subsumption_spec : forall {Γ A A'},
-    {{ Γ ⊢ A ⊆ A' }} ->
-    {{ ⊢ Γ }} /\ exists j, {{ Γ ⊢ A ≈ A' : Type@j }} \/ exists i i', i < i' /\ {{ Γ ⊢ Type@i ≈ A : Type@j }} /\ {{ Γ ⊢ A' ≈ Type@i' : Type@j }}.
-Proof.
-  intros * H.
-  dependent induction H; split; mauto 3.
-  - (* typ_subsumption_typ *)
-    eexists.
-    right.
-    do 2 eexists.
-    repeat split; revgoals; mauto.
-  - (* typ_subsumption_trans *)
-    destruct IHtyp_subsumption1 as [? [? [| [i1 [i1']]]]]; destruct IHtyp_subsumption2 as [? [? [| [i2 [i2']]]]];
-      destruct_conjs;
-      only 1: mautosolve 4;
-      eexists; right; do 2 eexists.
-    + (* left & right *)
-      split; [eassumption |].
-      solve [mauto using lift_exp_eq_max_left].
-    + (* right & left *)
-      split; [eassumption |].
-      solve [mauto using lift_exp_eq_max_right].
-    + exvar nat ltac:(fun j => assert {{ Γ ⊢ Type@i2 ≈ Type@i1' : Type@j }} by mauto).
-      replace i2 with i1' in * by mauto.
-      split; [etransitivity; revgoals; eassumption |].
-      mauto 4 using lift_exp_eq_max_left, lift_exp_eq_max_right.
-Qed.
 
-#[export]
-Hint Resolve typ_subsumption_spec : mcltt.
+(* Lemma typ_subsumption_spec : forall {Γ A A'}, *)
+(*     {{ Γ ⊢ A ⊆ A' }} -> *)
+(*     {{ ⊢ Γ }} /\ exists j, {{ Γ ⊢ A ≈ A' : Type@j }} \/ exists i i', i < i' /\ {{ Γ ⊢ Type@i ≈ A : Type@j }} /\ {{ Γ ⊢ A' ≈ Type@i' : Type@j }}. *)
+(* Proof. *)
+(*   intros * H. *)
+(*   dependent induction H; split; mauto 3. *)
+(*   - (* typ_subsumption_typ *) *)
+(*     eexists. *)
+(*     right. *)
+(*     do 2 eexists. *)
+(*     repeat split; revgoals; mauto. *)
+(*   - (* typ_subsumption_trans *) *)
+(*     destruct IHtyp_subsumption1 as [? [? [| [i1 [i1']]]]]; destruct IHtyp_subsumption2 as [? [? [| [i2 [i2']]]]]; *)
+(*       destruct_conjs; *)
+(*       only 1: mautosolve 4; *)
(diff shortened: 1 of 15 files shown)


HuStmpHrrr:
I think #121 needs to be split into multiple PRs otherwise this one will be too big. I am not sure what is a better way than accepting broken CI before this work is done.

Ailrun:
How about changing some proofs that require both model&syntax (such as the `completeness` theorem) to the `Admitted` one (while keep the original proof as a comment), and then just finish syntactic part?

theories/Core/Syntactic/System/Definitions.v
```diff
@@ -291,21 +289,57 @@ with wf_sub_eq : ctx -> ctx -> sub -> sub -> Prop :=
...
+  `( {{ ⊢ Γ }} ->
+     i < j ->
+     {{ Γ ⊢ Type@i ⊆ Type@j }} )
+| wf_subtyp_pi :
+  `( {{ Γ ⊢ A : Type@i }} ->
+     {{ Γ , A ⊢ B : Type@i }} ->
+     {{ Γ ⊢ A' : Type@i }} ->
+     {{ Γ , A' ⊢ B' : Type@i }} ->
+     {{ Γ ⊢ A' ⊆ A }} ->
+     {{ Γ , A' ⊢ B ⊆ B' }} ->
```
Ailrun, line 316: This (genuine) asymmetry bothers me a bit... By that, I mean, if we do not have argument subtyping `{{ Γ , A' ⊢ B ⊆ B' }}` and `{{ Γ , A ⊢ B ⊆ B' }}` would be equivalent, but now `{{ Γ , A ⊢ B ⊆ B' }}` doesn't even make sense (as `B'` may require the `A'`, not `A`, in its context).

I cannot come up with a real problem and this should be fine, but it feels a bit...
More concretely, something like this bothers me: after presupposition, we will have `{{ Γ, A' ⊢ B ⊆ B' }}` and get `exist i, {{ Γ, A' ⊢ B' : Type@i }}`. but then, this is not enough to remove `{{ Γ , A ⊢ B : Type@i }}` because this one has a stronger information. (`B` is still well-typed in a less informative context)

theories/Core/Syntactic/System/Definitions.v
```diff
@@ -291,21 +289,57 @@ with wf_sub_eq : ctx -> ctx -> sub -> sub -> Prop :=
...
+     {{ Γ ⊢ Π A B ⊆ Π A' B' }} )
+where "Γ ⊢ A ⊆ A'" := (wf_subtyping Γ A A') (in custom judg) : type_scope.
 
 Scheme wf_ctx_mut_ind := Induction for wf_ctx Sort Prop
-with wf_ctx_eq_mut_ind := Induction for wf_ctx_eq Sort Prop
+with wf_ctx_eq_mut_ind := Induction for wf_ctx_sub Sort Prop
```
Ailrun, line 321: Could you also rename the principle (to `wf_ctx_sub_mut_ind`)?

theories/Core/Syntactic/Presup.v
```diff
@@ -252,23 +259,16 @@ Proof with mautosolve 4.
...
+
+    (* presup_subtyping cases *)
+  - exists (max i i0); split; mauto 3 using lift_exp_max_left, lift_exp_max_right.
+  - exists (max (S i) (S j)); split; mauto 3 using lift_exp_max_left, lift_exp_max_right.
+  - mauto.
 Qed.
 
 #[export]
-Hint Resolve presup_exp presup_exp_eq presup_sub_eq : mcltt.
+Hint Resolve presup_exp presup_exp_eq presup_sub_eq presup_subtyping : mcltt.
```
HuStmpHrrr, line 270: @Ailrun are these hints even useful? I am thinking removing them.

Ailrun: Yes, I think some `mauto` cases indeed use them.

HuStmpHrrr: like what? why the tactics in Presup.v are not enough?

Ailrun: It is enough, but it requires 2 tactics (`gen_presup; mauto`). It should be fine to remove one modulo that increment.

Ailrun: But I suspect that that increment would be quite cumbersome. Why do you want to remove the hints?

HuStmpHrrr: I will try. my 5 cents bet on that they are not used at all.

Ailrun: There are something like this:
```
assert {{ |- \Gamma }} by mauto 3.
```
(Under context with `{{ \Gamma |- M : A }}`. How can you prove this without these hints?

HuStmpHrrr: ok. where do you find this? can't we call gen_presups? I think it's a good saturation lemma if you have lots of assertions like this.

Ailrun: Actually this is not a good example. There is `presup_exp_ctx` hint that already handles this case...
Ok, let's remove the hint and see what would happen. Could you do that in a different PR (so that we can see the impact on other parts of the proofs more easily)?

theories/Core/Syntactic/CoreInversions.v
```diff
@@ -118,15 +118,13 @@ Hint Resolve wf_sub_id_inversion : mcltt.
 
 Lemma wf_sub_weaken_inversion : forall {Γ Δ},
     {{ Γ ⊢s Wk : Δ }} ->
-    exists A, {{ ⊢ Γ ≈ Δ, A }}.
-Proof with mautosolve 4.
+    exists Γ' A, {{ ⊢ Γ ≈ Γ', A }} /\ {{ ⊢ Γ' ⊆ Δ }}.
```
Ailrun, line 121: Why not `{{ ⊢ Γ ⊆ Δ, A }}`?

theories/Core/Completeness/Consequences/Types.v
```diff
@@ -87,148 +87,149 @@ Hint Resolve is_typ_constr_and_exp_eq_nat_implies_eq_nat : mcltt.
 
 (* We cannot use this spec as the definition of [typ_subsumption] as
    then its transitivity requires [exp_eq_typ_implies_eq_level] or a similar semantic lemma *)
-Lemma typ_subsumption_spec : forall {Γ A A'},
```
HuStmpHrrr, line 90: I think many lemmas below are no longer true. We need to pick which ones are false.

theories/Core/Syntactic/CoreInversions.v
HuStmpHrrr: This won't be true. `A` might not be well-typed in `\Delta`.

HuStmpHrrr:
ok, syntactic part is ready

PR #124, opened 2024-06-22
working on algorithmic subtyping

HuStmpHrrr:
c.f. #121

### DIFF
--- a/theories/Algorithmic/Subtyping/Definitions.v
+++ b/theories/Algorithmic/Subtyping/Definitions.v
@@ -21,7 +21,7 @@ Inductive alg_subtyping_nf : nf -> nf -> Prop :=
     i <= j ->
     {{ ⊢anf Type@i ⊆ Type@j }}
 | asnf_pi : forall A B A' B',
-    {{ ⊢anf A' ⊆ A }} ->
+    A = A' ->
     {{ ⊢anf B ⊆ B' }} ->
     {{ ⊢anf Π A B ⊆ Π A' B' }}
 where "⊢anf M ⊆ N" := (alg_subtyping_nf M N) (in custom judg) : type_scope.
--- a/theories/Algorithmic/Subtyping/Definitions.v
+++ b/theories/Algorithmic/Subtyping/Definitions.v
@@ -4,7 +4,7 @@ From Mcltt Require Export Syntax.
 Import Syntax_Notations.
 
 Reserved Notation "Γ ⊢a A ⊆ A'" (in custom judg at level 80, Γ custom exp, A custom exp, A' custom exp).
-Reserved Notation "Γ ⊢anf A ⊆ A'" (in custom judg at level 80, Γ custom exp, A custom nf, A' custom nf).
+Reserved Notation "⊢anf A ⊆ A'" (in custom judg at level 80, A custom nf, A' custom nf).
 
 Definition not_univ_pi (A : nf) : Prop :=
   match A with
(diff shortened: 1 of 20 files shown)


HuStmpHrrr:
Some lemmas are left behind now, including the completeness proof.

HuStmpHrrr:
When proving subtyping in the Pi case, I noticed another problem. Now the algorithm first normalizes both sides and then just recurses on normal forms to decide subtyping. It looks right but the completeness proof is blocked. 

Consider `Pi A B` and `Pi A' B'`. The algorithm normalizes `B` to some `N` with an extended environment of `A`. The problem now, is the declarative subtyping can only confirm a subtyping relation between `B` and `B'` when the context is extended with `A'`. So the inductive argument doesn't go through.

I am not sure how to fix it. I hope it doesn't involve a complicated semantic setup. I somehow feel that the subtyping is just a fancy way of propagating cumulativity so it shouldn't be this difficult. I think the solution might just be to relate the normal forms of `B` obtained in different contexts, as we know `A'` is a subtype of `A`.

HuStmpHrrr:
I think the solution probably lies in the PER model of contexts, which we will have to express anyways. A lazyman's fix at this moment is to only say `A` is equivalent to `A'`, but as I pointed out in CoqClub, this induces some issue in the type theory.

theories/Algorithmic/Subtyping/Definitions.v
```diff
@@ -0,0 +1,40 @@
...
+
+
+Inductive alg_subtyping : ctx -> typ -> typ -> Prop :=
+| alg_subtyp_run :
+  forall Γ M N i j A B,
+    nbe Γ M {{{ Type@i }}} A ->
+    nbe Γ N {{{ Type@j }}} B ->
+    {{ ⊢anf A ⊆ B }} ->
+    {{ Γ ⊢a M ⊆ N }}
+where "Γ ⊢a M ⊆ N" := (alg_subtyping Γ M N) (in custom judg) : type_scope.
```
Ailrun, line 37: I'm not sure whether we want this separately. IMO, this should be more like a specification of subtyping, not a separate judgement.

Ailrun: Like, the above nf subtyping is actually "the" subtyping of normal forms, and this algorithmic subtyping is kind like of completeness&soundness of nbe but for subtyping.

Ailrun:
How about this algorithm?
1. For all G and G' where G is sub-context of G' normalize M in G and M' in G'.
2. decide normal form subtyping for W and W' (the normal forma obtained by the above step)
3. This gives us `G |- M sub M'`

This algorithm is definitely sound as we have `G |- M ~~ W`, `G |- W sub W'` and `G' |- W' ~~ M` (which can be "lowered" to `G |- W' ~~ M` and this the conclusion. The question is now more on completeness logrel for this I guess.

HuStmpHrrr:
> How about this algorithm?
> 
>     1. For given G and G' where G is sub-context of G' normalize M in G and M' in G'.
> 
>     2. decide normal form subtyping for W and W' (the normal forma obtained by the above step)
> 
>     3. This gives us `G |- M sub M'`
> 
> 
> This algorithm is definitely sound as we have `G |- M ~~ W`, `G |- W sub W'` and `G' |- W' ~~ M` (which can be "lowered" to `G |- W' ~~ M` and this the conclusion. The question is now more on completeness logrel for this I guess.

this is basically what is doing here. I am giving up on contra-variant subtyping. too many things fail on my end.

HuStmpHrrr:
should merge #127 first

theories/Core/Completeness/Consequences/Rules.v
```diff
@@ -0,0 +1,23 @@
+From Coq Require Import RelationClasses.
+From Mcltt Require Import Base LibTactics.
+From Mcltt.Core Require Import Completeness Completeness.FundamentalTheorem Completeness.LogicalRelation Semantic.Realizability PER.
+From Mcltt.Core Require Export SystemOpt.
+Import Domain_Notations.
+
+Lemma ctxeq_nbe_eq : forall Γ Γ' M A,
```
HuStmpHrrr, line 7: basically this is what we couldn't establish with contra-variant subtyping. we need to establish some relation between normal forms in different contexts.

Ailrun:
> this is basically what is doing here. I am giving up on contra-variant subtyping. too many things fail on my end.
The previous one is using same `Gamma`, not `Gamma` and `Gamma'` under a sub-context relation. But anyway, I'm all good for not having contravariant subtyping.

Ailrun (approved):

PR #125, opened 2024-06-22
[WIP] define subtyping for the PER model

HuStmpHrrr:
c.f. #121


Ailrun:
Yeah, I can see that a solution wouldn't be very simple (or clean). If you can put your exact approach it would be helpful for a more discussion.

HuStmpHrrr:
I have proved transitivity of subtyping of the PER model for the sake of it, just to show that the blocker for contra-variant subtyping is not coming from the semantics, but from the decidability of subtyping.

PR #127, opened 2024-06-27
remove contra-variant subtyping

HuStmpHrrr:
c.f. #121


HuStmpHrrr:
Let's merge with a merge commit just for this PR so I don't get blocked.

Ailrun:
oh oops I missed your last comment. My bad.

HuStmpHrrr:
Shouldn't be a big deal. Just need to resolve conflict but I don't think there's any.

PR #130, opened 2024-07-09
working on completeness rules for subtyping

HuStmpHrrr:
c.f. #121

### DIFF
--- a/theories/Core/Completeness/ContextCases.v
+++ b/theories/Core/Completeness/ContextCases.v
@@ -72,6 +72,6 @@ Proof.
   destruct H3 as [? [? []]].
   on_all_hyp: fun H => (eapply rel_ctx_extend' in H; destruct H).
   econstructor; mauto; intros.
-  deepexec H3 ltac:(fun H => destruct H as []).
+  deepexec H3 ltac:(fun H => destruct H as [? [? [? [? []]]]]).
   simplify_evals. eassumption.
 Qed.
--- a/theories/Core/Completeness/ContextCases.v
+++ b/theories/Core/Completeness/ContextCases.v
@@ -75,3 +75,7 @@ Proof.
   deepexec H3 ltac:(fun H => destruct H as [? [? [? [? []]]]]).
   simplify_evals. eassumption.
 Qed.
+
+
+#[export]
+Hint Resolve wf_ctx_sub_empty' wf_ctx_sub_extend' : mcltt.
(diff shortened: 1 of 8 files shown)


HuStmpHrrr:
@Ailrun could you please take a look at the function case and see if it can use some lemmas you have proved for the function cases for the completeness proof, after the popl deadline?

theories/Core/Completeness/SubtypingCases.v
```diff
@@ -0,0 +1,179 @@
...
+  intros.
+  do 2 eexists; repeat split;
+  econstructor; mauto 2.
+  - apply per_univ_elem_core_univ'; [lia | reflexivity].
+  - apply per_univ_elem_core_univ'; [lia | reflexivity].
+  - econstructor; lia.
+Qed.
+
+
+Lemma wf_subtyp_pi' : forall Γ A A' B B' i,
```
HuStmpHrrr, line 87: @Ailrun Idk if you have better idea in cleaning this proof up. maybe you could give a shot after merging this PR.

HuStmpHrrr:
Let's merge this PR first, and then we have another PR just to fix up the completeness proof.

theories/Core/Completeness/SubtypingCases.v
Ailrun: Sure

theories/Core/Completeness/ContextCases.v
```diff
@@ -41,5 +41,41 @@ Proof with intuition.
...
+  on_all_hyp: fun H => (eapply rel_ctx_extend' in H; destruct H).
+  econstructor; mauto; intros.
+  deepexec H3 ltac:(fun H => destruct H as [? [? [? [? []]]]]).
+  simplify_evals. eassumption.
+Qed.
+
+
 #[export]
-Hint Resolve rel_ctx_extend : mcltt.
+Hint Resolve wf_ctx_sub_empty' wf_ctx_sub_extend' : mcltt.
```
Ailrun, line 81: In my opinion, these names are better suited for other things, such as alternative constructo s (or helper lemmas) for the corresponding constructors. That's why I used different names for other completeness cases. Are there other names for these?

HuStmpHrrr: do you have a naming pattern?

Ailrun: I used `<relation type>_<constructor suffix>`, like when the type is `rel_ctx` and the constructor was `wf_ctx_extend`, I combined `rel_ctx` and `extend` to get `rel_ctx_extend`. Does this work for these cases as well?

HuStmpHrrr: I will see

HuStmpHrrr: ok I think it's good.

Ailrun (approved):

Issue #131, opened 2024-07-12
Make Gluing Model Unary


Ailrun:
C.C. @HuStmpHrrr 

In the gluing model definition of
```coq
Inductive glu_univ_elem_core : glu_typ_pred -> glu_exp_pred -> domain -> domain -> Prop
```
why do we need two `domain` types? In the Agda mechanization, sure, we used `A ≈ B ∈ 𝕌 j` (actually even there, why doesn't `A ∈′ 𝕌 j` work? Our `transport` or other lemmas there are all one-sided after all), but here we repeat the structure of `A ≈ B ∈ 𝕌 j` in the model anyway, and what we need is relating one syntactic term with one domain term.

Ailrun:
I got this question while defining the gluing model for substitutions. Even in the Agda mechanization, we used a one-sided definition for the model instead of a binary one.

HuStmpHrrr:
I think in principle the gluing model is ok to be unary. try to change it to unary. I think it will just work.

Ailrun:
Yeah, I will try.

HuStmpHrrr:
I think that will probably work out better for the subtyping relation. I think you can also try to define subtyping for gluing. it might be a bit twisted.

Ailrun:
#132 resolves this.

PR #132, opened 2024-07-12
Some updates on soundness

Ailrun:


theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -10,13 +10,23 @@ Global Open Scope predicate_scope.
 
 Generalizable All Variables.
 
-Notation "'typ_pred'" := (predicate (Tcons ctx (Tcons typ Tnil))).
-Notation "'typ_pred_equivalence'" := (@predicate_equivalence (Tcons ctx (Tcons typ Tnil))).
-Notation "'glu_pred'" := (predicate (Tcons ctx (Tcons exp (Tcons typ (Tcons domain Tnil))))).
-Notation "'glu_pred_equivalence'" := (@predicate_equivalence (Tcons ctx (Tcons exp (Tcons typ (Tcons domain Tnil))))).
```
Ailrun, line 16: I change the nomenclature for these names so that `glu_sub_pred` fits better into them.

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -10,13 +10,23 @@ Global Open Scope predicate_scope.
...
+Notation "'glu_exp_pred'" := (predicate glu_exp_pred_args).
+Notation "'glu_exp_pred_equivalence'" := (@predicate_equivalence glu_exp_pred_args).
+Notation "Γ ⊢ M : A ® m ∈ R" := (R Γ M A m : (Prop : Type)) (in custom judg at level 80, Γ custom exp, M custom exp, A custom exp, m custom domain, R constr).
+
+Notation "'glu_sub_pred_args'" := (Tcons ctx (Tcons sub (Tcons env Tnil))).
+Notation "'glu_sub_pred'" := (predicate glu_sub_pred_args).
+Notation "'glu_sub_pred_equivalence'" := (@predicate_equivalence glu_sub_pred_args).
+Notation "Γ ⊢s σ ® ρ ∈ R" := (R Γ σ ρ : Prop) (in custom judg at level 80, Γ custom exp, σ custom exp, ρ custom domain, R constr).
+
+Notation "'Glu' a ≈ b ∈ R ↘ P ↘ El" := (R P El a b : Prop) (in custom judg at level 90, a custom domain, b custom domain, R constr, P constr, El constr).
```
Ailrun, line 28: C.f. #131.

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -28,209 +38,206 @@ Inductive glu_nat : ctx -> exp -> domain -> Prop :=
...
-Arguments neut_typ_pred i C Γ M/.
+            (forall Δ σ V, {{ Δ ⊢w σ : Γ }} -> {{ Rne C in length Δ ↘ V }} -> {{ Δ ⊢ M[σ] ≈ V : Type@i }}).
+Arguments neut_glu_typ_pred i C Γ M/.
 
-Inductive neut_glu_pred i C : glu_pred :=
-| ngp_make : forall Γ m M A c,
-    neut_typ_pred i C Γ M ->
-    per_bot c c ->
+Inductive neut_glu_exp_pred i C : glu_exp_pred :=
+| mk_neut_glu_exp_pred : forall Γ m M A c,
```
Ailrun, line 59: I change the names of constructors like this as other one-constructor types use the `mk` prefix.

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -28,209 +38,206 @@ Inductive glu_nat : ctx -> exp -> domain -> Prop :=
...
-Equations glu_univ_elem (i : nat) : typ_pred -> glu_pred -> domain -> domain -> Prop by wf i :=
-| i => glu_univ_elem_core i (fun j lt_j_i A Γ M => exists P El, glu_univ_elem j P El A A /\ P Γ M).
-
-
-Definition glu_univ (i : nat) (A : domain) : typ_pred :=
-  fun Γ M => exists P El, glu_univ_elem i P El A A /\ P Γ M.
-Arguments glu_univ i A Γ M/.
+Definition glu_univ_typ (i : nat) (A : domain) (B : domain) : glu_typ_pred :=
+  fun Γ M => exists P El, {{ Glu A ≈ B ∈ glu_univ_elem i ↘ P ↘ El }} /\ {{ Γ ⊢ M ® P }}.
+Arguments glu_univ_typ i A B Γ M/.
```
Ailrun, line 173: I change this one to take 2 types so that we can use it for `glu_neut` and `glu_norm`, but if we take the path of #131, I will roll this back (while keeping the notations).

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -28,209 +38,206 @@ Inductive glu_nat : ctx -> exp -> domain -> Prop :=
...
 
-  Definition univ_glu_pred' {j} (lt_j_i : j < i) : glu_pred :=
+  Definition univ_glu_exp_pred' {j} (lt_j_i : j < i) : glu_exp_pred :=
     fun Γ m M A =>
-      {{ Γ ⊢ m : M }} /\ {{ Γ ⊢ M ≈ Type@j : Type@i }} /\
-        per_univ j A A /\
-        glu_univ_typ_rec lt_j_i A Γ m.
+      {{ Γ ⊢ m : M }} /\
+        {{ Γ ⊢ M ≈ Type@j : Type@i }} /\
+        {{ Γ ⊢ m ® glu_univ_typ_rec lt_j_i A }}.
```
Ailrun, line 118: I remove `per_univ` as it is implied (and does not block the proof of `glu_univ -> per_univ` lemma).

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -246,45 +253,121 @@ Section GluingInduction.
...
+| mk_glu_typ : forall P El,
     {{ Γ ⊢ M : Type@i }} ->
-    glu_univ_elem i P El A B ->
-    per_top_typ A B ->
-    (forall Δ σ a, {{ Δ ⊢w σ : Γ }} -> {{ Rtyp A in length Δ ↘ a }} -> {{ Δ ⊢ M [ σ ] ≈  a : Type@i }}) ->
-    glu_typ i Γ M A B.
-
+    {{ Glu A ≈ B ∈ glu_univ_elem i ↘ P ↘ El }} ->
+    (forall Δ σ a, {{ Δ ⊢w σ : Γ }} -> {{ Rtyp A in length Δ ↘ a }} -> {{ Δ ⊢ M[σ] ≈ a : Type@i }}) ->
+    {{ Γ ⊢ M ® glu_typ i A B }}.
```
Ailrun, line 279: Here, I remove `per_top_typ` as `glu_univ_elem` implies that (through `per_univ_elem` using the semantic realizability).

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -246,45 +253,121 @@ Section GluingInduction.
...
+(* The parameters are ordered differently from the Agda version
+   so that we can return "glu_sub_pred". *)
+Variant cons_glu_sub_pred i Γ A (TSb : glu_sub_pred) : glu_sub_pred :=
+| mk_cons_glu_sub_pred :
+  `{ forall P El,
+        {{ Δ ⊢s σ : Γ, A }} ->
+        (* Do we need the following argument?
+           In other words, is it possible to prove that "TSb" respects
+           wf_sub_eq without the following condition? *)
+        {{ Δ ⊢s Wk ∘ σ ≈ Wk∘σ : Γ }} ->
```
Ailrun, line 314: Not sure we need this or not.

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -6,15 +6,27 @@ From Mcltt.Core.Soundness Require Import LogicalRelation.Definitions.
 From Mcltt.Core.Soundness Require Export Weakening.Lemmas.
 Import Domain_Notations.
 
+Lemma pi_glu_exp_pred_pi_glu_typ_pred : forall i IR IP IEl (OP : forall c c' (equiv_c_c' : {{ Dom c ≈ c' ∈ IR }}), glu_typ_pred) elem_rel OEl Γ m M a,
+    {{ Γ ⊢ m : M ® a ∈ pi_glu_exp_pred i IR IP IEl elem_rel OEl }} ->
+    (forall Δ m' M' b c c' (equiv_c_c' : {{ Dom c ≈ c' ∈ IR }}),
+        {{ Δ ⊢ m' : M' ® b ∈ OEl _ _ equiv_c_c' }} ->
+        {{ Δ ⊢ M' ® OP _ _ equiv_c_c' }}) ->
+    {{ Γ ⊢ M ® pi_glu_typ_pred i IR IP IEl OP }}.
```
Ailrun, line 14: This one seems relevant to some lemmas below, but is never actually used. Maybe there is some duplication of proof?

theories/Core/Soundness/LogicalRelation/Definitions.v
HuStmpHrrr: both sides look identical to me. what's the catch here?

Ailrun: Oh right... my mistake. I meant to use a new variable for `Wk∘σ`, which was the way how we defined this in Agda. If I remember correctly, in the Agda case, we defined it like so to make some proof easier, not because it is absolutely necessary, but here I'm not sure
1. Whether it is really unnecessary
2. Even when it is unnecessary, whether it reduces proof much or not.

HuStmpHrrr: this variable is to pass onto the recursive case for the gluing model. if it appears somewhere, it should appear within the inductive definition, probably not here.

theories/Core/Completeness/ContextCases.v
```diff
@@ -1,6 +1,6 @@
 From Coq Require Import Morphisms_Relations.
 From Mcltt Require Import Base LibTactics.
-From Mcltt.Core Require Import Completeness.LogicalRelation Completeness.UniverseCases.
+From Mcltt.Core Require Import Evaluation Completeness.LogicalRelation Completeness.UniverseCases.
```
HuStmpHrrr, line 3: Do you need anything from completeness? is this a merge branch?

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -10,281 +10,358 @@ Global Open Scope predicate_scope.
...
 Generalizable All Variables.
 
-Notation "'typ_pred'" := (predicate (Tcons ctx (Tcons typ Tnil))).
-Notation "'typ_pred_equivalence'" := (@predicate_equivalence (Tcons ctx (Tcons typ Tnil))).
-Notation "'glu_pred'" := (predicate (Tcons ctx (Tcons exp (Tcons typ (Tcons domain Tnil))))).
-Notation "'glu_pred_equivalence'" := (@predicate_equivalence (Tcons ctx (Tcons exp (Tcons typ (Tcons domain Tnil))))).
+Notation "'glu_typ_pred_args'" := (Tcons ctx (Tcons typ Tnil)).
+Notation "'glu_typ_pred'" := (predicate glu_typ_pred_args).
+Notation "'glu_typ_pred_equivalence'" := (@predicate_equivalence glu_typ_pred_args).
+Notation "Γ ⊢ A ® R" := ((R Γ A : Prop) : Prop) (in custom judg at level 80, Γ custom exp, A custom exp, R constr).
```
HuStmpHrrr, line 16: why two layers of Prop?

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -10,281 +10,358 @@ Global Open Scope predicate_scope.
...
 
-Definition univ_typ_pred j i : typ_pred := fun Γ T => {{ Γ ⊢ T ≈ Type@j :  Type@i }}.
-Arguments univ_typ_pred j i Γ T/.
+Notation "'glu_exp_pred_args'" := (Tcons ctx (Tcons exp (Tcons typ (Tcons domain Tnil)))).
+Notation "'glu_exp_pred'" := (predicate glu_exp_pred_args).
+Notation "'glu_exp_pred_equivalence'" := (@predicate_equivalence glu_exp_pred_args).
+Notation "Γ ⊢ M : A ® m ∈ R" := (R Γ M A m : (Prop : Type)) (in custom judg at level 80, Γ custom exp, M custom exp, A custom exp, m custom domain, R constr).
```
HuStmpHrrr, line 21: why Prop:Type?

theories/Core/Soundness/LogicalRelation/Definitions.v
Ailrun: Just to distinguish it with other judgments.

theories/Core/Soundness/LogicalRelation/Definitions.v
Ailrun: Like above. `PER.Definitions` already uses this trick.

HuStmpHrrr:
make a cut somewhere and merge this PR before it gets too big. let's address some issues first.

theories/Core/Completeness/ContextCases.v
Ailrun: It's from merge

Ailrun:
Sure, let's merge this.

PR #137, opened 2024-07-24
Refactor gen rewritings

### LINKED ISSUE #135: Make gluing model works with generalized rewriting
~~They are defined as `binary_relation`, which blocks all `Hint Extern`s in https://github.com/coq/coq/blob/38d6695084c9d38b5af7258a85702a1d01a49dce/theories/Classes/Morphisms.v due to
https://github.com/coq/coq/blob/38d6695084c9d38b5af7258a85702a1d01a49dce/theories/Classes/Morphisms.v#L37-L53
Note that all of these uses `relation`, not `binary_relation`. Once unfolded, they are identical, but Ltac does not know the unfolded structure and typeclass resolution fails.~~

~~Without those typeclasses, it is hard to use the generalized rewriting, which helps a lot when dealing with functionality of the gluing model.~~

It turns out one possible root cause is opaque `arrows`. Generalized rewriting works with `Typeclasses Transparent arrows`.

Ailrun:
Resolve #135.

### DIFF
--- a/theories/Core/Semantic/PER/Lemmas.v
+++ b/theories/Core/Semantic/PER/Lemmas.v
@@ -10,6 +10,12 @@ Proof.
   split; intros []; econstructor; unfold Symmetric, Transitive in *; intuition.
 Qed.
 
+#[export]
+Instance subrelation_relation_equivalence {A} : relation_equivalence ~> pointwise_relation A (pointwise_relation A iff).
+Proof.
+  intros R R' HRR' x x'; intuition.
+Qed.
+
 Add Parametric Morphism R0 `(R0_morphism : Proper _ ((@relation_equivalence domain) ==> (@relation_equivalence domain)) R0) A p A' p' : (rel_mod_eval R0 A p A' p')
     with signature (@relation_equivalence domain) ==> iff as rel_mod_eval_morphism.
 Proof.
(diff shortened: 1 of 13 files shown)


Ailrun:
I have no idea why
https://github.com/Beluga-lang/McLTT/blob/1642d65a50606581210307fb4c3b766f5f17205c/theories/Core/Soundness/LogicalRelation/Lemmas.v#L399
loops forever.

I really think it would be much easier if we just define our own equivalence relations instead of that `predicate_equivalence` thing.

theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -10,6 +10,12 @@ Proof.
   split; intros []; econstructor; unfold Symmetric, Transitive in *; intuition.
 Qed.
 
+#[export]
```
HuStmpHrrr, line 13: this should go under lib tactics

theories/LibTactics.v
```diff
@@ -407,3 +407,16 @@ Qed.
...
+
+(* Helper Instances for Generalized Rewriting *)
+Add Parametric Morphism A : PER
+    with signature (@relation_equivalence A) ==> iff as PER_morphism.
+Proof.
+  split; intros []; econstructor; unfold Symmetric, Transitive in *; intuition.
+Qed.
+
+#[export]
+Instance subrelation_relation_equivalence {A} : relation_equivalence ~> pointwise_relation A (pointwise_relation A iff).
```
HuStmpHrrr, line 419: I think `~>` might fail in the CI. this notation doesn't exist here. say `subrelation` instead.

Ailrun: Yeah probably that was the case

Issue #136, opened 2024-07-23
Soundness and Logical Relations


HuStmpHrrr:
@Ailrun Let's split work in this issue. List and pick stuff that you want to work on so we don't overlap.

- [x] subtyping
  - #151
  - #153
  - #157
- [x] cumulativity
  - #150
- [x] realizability
  - #140  
- [ ] typing rules
  - #113 (Roadmap issue)

Ailrun:
I will first fix #135.

HuStmpHrrr:
I will look into realizability. subtyping is gonna be interesting. @Ailrun do you want to handle it? or do you want to swap.

Ailrun:
By subtyping, you mean subtyping cases? I didn't get it when you said subtyping logrel for soundness, as I expected something like
https://gitlab.com/JasonHuZS/practice/-/blob/master/MLTT/Soundness/Terms.agda#L22-26
(that using syntactic judgement) or similar.

HuStmpHrrr:
there should be lemmas for subtyping in the logical relations too. it should quantify what happens to the logical relation when there is a subtyping in the domain.

Ailrun:
You mean something like
```
Lemma glu_univ_elem_typ_subtyping : forall {i a a' P P' El El' Γ A},
    {{ Sub a <: a' at i }} ->
    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
    {{ DG a' ∈ glu_univ_elem i ↘ P' ↘ El' }} ->
    P Γ A ->
    exists A', {{ Γ ⊢ A ⊆ A' }} /\ P' Γ A'.
```

HuStmpHrrr:
It should be universal over A', but otherwise I think this should be the theorem. Not sure if there's more.

Ailrun:
How it can be universal in `A'`? I mean, `A'` should be a specific type that corresponds to `a'`, not any supertype of `A`.

HuStmpHrrr:
put `P' Γ A'` in the implication too. you just prove A is a subtype of A'.

Ailrun:
Then what would that provide? For subtyping cases we need if `El G M A m` then `El' G M A' m` for some supertype `A'`

HuStmpHrrr:
this one will also prove the previous lemma as part of it. proving this one should also work.

Ailrun:
It seems like the subtyping lemmas require realizability. In the following lemma,
```
[...]

PR #140, opened 2024-07-26
working on realizability

HuStmpHrrr:
c.f. #136 #121

### DIFF
--- a/theories/Core/Soundness/LogicalRelation/Definitions.v
+++ b/theories/Core/Soundness/LogicalRelation/Definitions.v
@@ -252,28 +252,38 @@ Section GluingInduction.
   Qed.
 End GluingInduction.
 
-Inductive glu_neut i A Γ M m c : Prop :=
-| mk_glu_neut :
-    {{ Γ ⊢ m : M }} ->
-    {{ Γ ⊢ M ® glu_univ_typ i A }} ->
-    {{ Dom c ≈ c ∈ per_bot }} ->
-    (forall Δ σ a, {{ Δ ⊢w σ : Γ }} -> {{ Rne c in length Δ ↘ a }} -> {{ Δ ⊢ m[σ] ≈ a : M[σ] }}) ->
-    {{ Γ ⊢ m : M ® c ∈ glu_neut i A }}.
-
-Inductive glu_norm i A Γ M m a : Prop :=
-| mk_glu_norm :
-    {{ Γ ⊢ m : M }} ->
-    {{ Γ ⊢ M ® glu_univ_typ i A }} ->
-    {{ Dom ⇓ A a ≈ ⇓ A a ∈ per_top }} ->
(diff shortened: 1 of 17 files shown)


theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -73,6 +73,7 @@ Inductive pi_glu_typ_pred i
   (OP : forall c (equiv_c : {{ Dom c ≈ c ∈ IR }}), glu_typ_pred) : glu_typ_pred :=
 | mk_pi_glu_typ_pred :
     `{ {{ Γ ⊢ M ≈ Π IT OT : Type@i }} ->
+       {{ Γ ⊢ IT : Type@i }} ->
```
Ailrun, line 76: Why do we need these (the well-formednesses of `IT` and `OT`)? They are derivable from a presupposition of the previous assumption, i.e. the well-formedness of Pi type. Do we need these for some induction-related reason?

HuStmpHrrr: They can be derived but it's a bit annoying to not have them. Extracting them all the time from the logical relations sounds repetitive. They are almost always needed when given as premises. For conclusion, I think we can provide a simplified constructor as always.

HuStmpHrrr: In fact, adding this premise doesn't change the proofs of the lemmas at all. I think automation just crushes it. It's a pure gain.

HuStmpHrrr:
This PR is good.

theories/Core/Syntactic/SystemOpt.v
```diff
@@ -285,3 +285,96 @@ Qed.
...
+    {{ Γ ⊢ ℕ[σ] ≈ ℕ : Type@i }}.
+Proof.
+  intros.
+  eapply wf_exp_eq_subtyp with (A := {{{ Type@0 }}}).
+  - mauto 3.
+  - gen_presups.
+    eapply wf_subtyping_ge;
+      eauto.
+    lia.
+Qed.
```
Ailrun, line 301:
```suggestion
Proof.
  intros.
  assert (0 <= i) by lia.
  mauto.
Qed.
```
is enough

Ailrun: (You may be even able to replace `mauto` with `mauto 4` or so)

theories/Core/Syntactic/SystemOpt.v
```diff
@@ -285,3 +285,96 @@ Qed.
...
+Lemma wf_subtyping_subst : forall Δ A B,
+    {{ Δ ⊢ A ⊆ B }} ->
+    forall Γ σ,
+      {{ Γ ⊢s σ : Δ }} ->
+      {{ Γ ⊢ A [σ] ⊆ B[σ] }}.
+Proof.
+  intros; mauto 2 using wf_subtyping_subst_eq.
+Qed.
+#[export]
+ Hint Resolve wf_subtyping_subst_eq wf_subtyping_subst : mcltt.
```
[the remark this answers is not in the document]
HuStmpHrrr: ok moved over

PR #149, opened 2024-07-31
implement monotonicity

### LINKED ISSUE #143: Finalize realizability once monotonicity is proved
c.f. https://github.com/Beluga-lang/McLTT/pull/140/files#diff-4cd5bf480f7ea77d50965f1b367793cbb1f7ffe428ce9def5998736e8e6e5a0fR129

HuStmpHrrr:
Closes #143

### DIFF
--- a/theories/Core/Soundness/LogicalRelation/Lemmas.v
+++ b/theories/Core/Soundness/LogicalRelation/Lemmas.v
@@ -135,7 +135,9 @@ Proof.
     simpl_glu_rel; repeat split; mauto.
 
   intros.
-  assert {{ Δ ⊢ M[σ] ≈ V : Type@i }}; mauto.
+  symmetry in H3.
+  saturate_weakening_escape.
+  bulky_rewrite.
 Qed.
 
 Add Parametric Morphism i P El A (H : glu_univ_elem i P El A) Γ : (El Γ)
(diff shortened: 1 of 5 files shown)


theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -135,7 +135,9 @@ Proof.
     simpl_glu_rel; repeat split; mauto.
 
   intros.
-  assert {{ Δ ⊢ M[σ] ≈ V : Type@i }}; mauto.
+  symmetry in H3.
```
Ailrun, line 138: Is it possible to avoid this explicit `H3`? This makes proof maintenance quite hard; if we add/remove some assumptions (possibly because we refactor some tactics), `symmetry` applies to any PER, so the proof refactorer should find the working hypothesis with a fairly restricted information.

HuStmpHrrr: ok I just put it back

theories/Core/Syntactic/SystemOpt.v
```diff
@@ -301,3 +301,16 @@ Qed.
...
+    {{ Γ ⊢s σ : Δ }} ->
+    i < j ->
+    {{ Γ ⊢ Type@i[σ] ≈ Type@i : Type@j }}.
+Proof. mauto 3. Qed.
+
+#[export]
+ Hint Resolve wf_exp_eq_typ_sub' : mcltt.
+
+#[export]
+Hint Rewrite -> wf_exp_eq_typ_sub' using solve [try lia; mauto 3] : mcltt.
```
Ailrun, line 316: If you expect `lia` to solve some and `mauto` to solve others, you can use `solve [lia | mauto 3]` (`solve` will try `|`-separated tactics in order)

Ailrun (approved):
Other than the above minor nitpicking, LGTM.

theories/Core/Syntactic/SystemOpt.v
HuStmpHrrr: alright

PR #150, opened 2024-07-31
Gluing Cumulativity

### LINKED ISSUE #146: Cumulativity should relate input and output relations
https://github.com/Beluga-lang/McLTT/blob/86f48d948352683b1c6c28413db5d1d2894c5d0d/theories/Core/Soundness/LogicalRelation/Lemmas.v#L623-L627

P and El should not change. This is how it is in Agda: https://hustmphrrr.github.io/mech-type-theories/MLTT.Soundness.Cumulativity.html

Ailrun:
Resolves #146. C.f. #136.

### DIFF
--- a/theories/Core/Soundness/Cumulativity.v
+++ b/theories/Core/Soundness/Cumulativity.v
@@ -1,0 +1,165 @@
+From Coq Require Import Morphisms Morphisms_Prop Morphisms_Relations Relation_Definitions RelationClasses.
+
+From Mcltt Require Import Base LibTactics.
+From Mcltt.Core.Syntactic Require Import Corollaries.
+From Mcltt.Core.Semantic Require Import Realizability.
+From Mcltt.Core.Soundness Require Import Realizability.
+From Mcltt.Core.Soundness Require Export LogicalRelation.
+Import Domain_Notations.
+
+Section glu_univ_elem_cumulativity.
+  #[local]
+  Lemma glu_univ_elem_cumulativity_ge : forall {i j a P P' El El'},
+      i <= j ->
+      {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+      {{ DG a ∈ glu_univ_elem j ↘ P' ↘ El' }} ->
+      (forall Γ A, {{ Γ ⊢ A ® P }} -> {{ Γ ⊢ A ® P' }}) /\
+        (forall Γ A M m, {{ Γ ⊢ M : A ® m ∈ El }} -> {{ Γ ⊢ M : A ® m ∈ El' }}) /\
+        (forall Γ A M m, {{ Γ ⊢ A ® P }} -> {{ Γ ⊢ M : A ® m ∈ El' }} -> {{ Γ ⊢ M : A ® m ∈ El }}).
+  Proof.
+    simpl.
+    intros * Hge Hglu Hglu'. gen El' P' j.
+    induction Hglu using glu_univ_elem_ind; repeat split; intros;
+      try assert {{ DF a ≈ a ∈ per_univ_elem j ↘ in_rel }} by mauto;
+      invert_glu_univ_elem Hglu';
+      handle_functional_glu_univ_elem;
+      simpl in *;
+      try solve [repeat split; intros; destruct_conjs; mauto 3 | intuition; mauto 4].
+
+    - rename x into IP'.
+      rename x0 into IEl'.
+      rename x1 into OP'.
+      rename x2 into OEl'.
+      destruct_by_head pi_glu_typ_pred.
+      econstructor; intros; mauto 4.
+      + assert {{ Δ ⊢ IT[σ] ® IP }} by mauto.
+        assert (forall Γ A, {{ Γ ⊢ A ® IP }} -> {{ Γ ⊢ A ® IP' }}) by (eapply proj1; mauto).
+        mauto.
+      + rename a0 into c.
+        rename equiv_a into equiv_c.
+        match_by_head per_univ_elem ltac:(fun H => directed invert_per_univ_elem H).
+        apply_relation_equivalence.
+        destruct_rel_mod_eval.
+        handle_per_univ_elem_irrel.
+        assert (forall Γ A, {{ Γ ⊢ A ® OP c equiv_c }} -> {{ Γ ⊢ A ® OP' c equiv_c }}) by (eapply proj1; mauto).
+        enough {{ Δ ⊢ OT[σ,,m] ® OP c equiv_c }} by mauto.
+        enough {{ Δ ⊢ m : IT[σ] ® c ∈ IEl }} by mauto.
+        eapply IHHglu; mauto.
+    - rename x into IP'.
+      rename x0 into IEl'.
+      rename x1 into OP'.
+      rename x2 into OEl'.
+      destruct_by_head pi_glu_exp_pred.
+      handle_per_univ_elem_irrel.
+      econstructor; intros; mauto 4.
+      + assert (forall Γ A, {{ Γ ⊢ A ® IP }} -> {{ Γ ⊢ A ® IP' }}) by (eapply proj1; mauto).
+        mauto.
+      + rename b into c.
+        rename equiv_b into equiv_c.
+        match_by_head per_univ_elem ltac:(fun H => directed invert_per_univ_elem H).
+        handle_per_univ_elem_irrel.
+        destruct_rel_mod_eval.
+        destruct_rel_mod_app.
+        handle_per_univ_elem_irrel.
+        eexists; split; mauto.
+        assert (forall Γ A M m, {{ Γ ⊢ M : A ® m ∈ OEl c equiv_c }} -> {{ Γ ⊢ M : A ® m ∈ OEl' c equiv_c }}) by (eapply proj1, proj2; mauto).
+        enough {{ Δ ⊢ m[σ] m' : OT[σ,,m'] ® fa ∈ OEl c equiv_c }} by mauto.
+        assert {{ Δ ⊢ m' : IT[σ] ® c ∈ IEl }} by (eapply IHHglu; mauto).
+        assert (exists ac, {{ $| a0 & c |↘ ac }} /\ {{ Δ ⊢ m[σ] m' : OT[σ,,m'] ® ac ∈ OEl c equiv_c }}) by mauto.
+        destruct_conjs.
+        functional_eval_rewrite_clear.
+        mauto.
+    - rename x into IP'.
+      rename x0 into IEl'.
+      rename x1 into OP'.
+      rename x2 into OEl'.
+      destruct_by_head pi_glu_typ_pred.
+      destruct_by_head pi_glu_exp_pred.
+      handle_per_univ_elem_irrel.
+      econstructor; intros; mauto.
+      rename b into c.
+      rename equiv_b into equiv_c.
+      match_by_head per_univ_elem ltac:(fun H => directed invert_per_univ_elem H).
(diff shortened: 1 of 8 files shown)


theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -799,6 +771,34 @@ Proof.
       * mauto 3.
 Qed.
 
+#[export]
```
HuStmpHrrr, line 774: why do we need this? I am most definitely sure there is a more elegant way to deal with it

theories/Core/Syntactic/SystemOpt.v
```diff
@@ -78,6 +92,35 @@ Hint Resolve wf_app' : mcltt.
...
+    {{ Γ ⊢ Type@i[σ] ≈ Type@i : Type@j }}.
+Proof. mauto 3. Qed.
+
+#[export]
+Hint Resolve wf_exp_eq_typ_sub' : mcltt.
+
+#[export]
+Hint Rewrite -> wf_exp_eq_typ_sub' using solve [lia | mauto 3] : mcltt.
+
+Corollary wf_exp_eq_nat_sub' : forall Γ σ Δ i,
```
HuStmpHrrr, line 107: this lemma already exists https://github.com/Beluga-lang/McLTT/blob/main/theories/Core/Syntactic/SystemOpt.v#L290

theories/Core/Soundness/LogicalRelation/Definitions.v
```diff
@@ -24,10 +24,10 @@ Notation "Γ ⊢ M : A ® m ∈ R" := (R Γ A M m : (Prop : (Type : Type))) (in
 Notation "'glu_sub_pred_args'" := (Tcons ctx (Tcons sub (Tcons env Tnil))).
 Notation "'glu_sub_pred'" := (predicate glu_sub_pred_args).
 Notation "'glu_sub_pred_equivalence'" := (@predicate_equivalence glu_sub_pred_args) (only parsing).
-Notation "Γ ⊢s σ ® ρ ∈ R" := (R Γ σ ρ : (Prop : (Type : Type))) (in custom judg at level 80, Γ custom exp, σ custom exp, ρ custom domain, R constr).
+Notation "Γ ⊢s σ ® ρ ∈ R" := ((R Γ σ ρ : Prop) : (Prop : (Type : Type))) (in custom judg at level 80, Γ custom exp, σ custom exp, ρ custom domain, R constr).
```
HuStmpHrrr, line 27: what motivated the change?

HuStmpHrrr: I even think that these notations should be parse only, so there isn't even need to do simpl. we always work with the raw form during proving anyways.

theories/Core/Soundness/Cumulativity.v
```diff
@@ -0,0 +1,165 @@
...
+From Mcltt.Core.Soundness Require Export LogicalRelation.
+Import Domain_Notations.
+
+Section glu_univ_elem_cumulativity.
+  #[local]
+  Lemma glu_univ_elem_cumulativity_ge : forall {i j a P P' El El'},
+      i <= j ->
+      {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+      {{ DG a ∈ glu_univ_elem j ↘ P' ↘ El' }} ->
+      (forall Γ A, {{ Γ ⊢ A ® P }} -> {{ Γ ⊢ A ® P' }}) /\
```
HuStmpHrrr, line 16: I suspect that you need all those morphisms below because there is a mismatch between the relation and this unfolded universal quantifications here. but there already is a notation defined in Coq: https://github.com/coq/coq/blob/master/theories/Classes/RelationClasses.v#L401

you can simply write `P -∙> P'`. this should fix your rewrite problems.

theories/Core/Soundness/LogicalRelation/Definitions.v
Ailrun: I'd like to see them whenever possible. For example, in `assert`ed statements, I can see them in the proof state.

theories/Core/Syntactic/SystemOpt.v
Ailrun: This one is just a renamed version of that one so that its naming is more consistent.

theories/Core/Soundness/LogicalRelation/Lemmas.v
Ailrun: Well, that's what we did for PER model. I also don't like that one but you said "it's just an instance".

theories/Core/Soundness/LogicalRelation/Definitions.v
HuStmpHrrr: but why do you keep adding more and more annotations? you don't seem to add any notations?

theories/Core/Soundness/Cumulativity.v
Ailrun: No, this one doesn't do anything with the rewriting. I rewrite things after introducing all of the assumptions anyway.

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: I thought you just need to add one general instance?

theories/Core/Soundness/LogicalRelation/Definitions.v
Ailrun: I saw some mismatch in the printing, so I added them to avoid the mismatch

theories/Core/Soundness/LogicalRelation/Lemmas.v
Ailrun: What would you place for `@pointwise_relation ... (@pointwise_relation ... (...)))`? The generalized version is just really `predicate_equivalence`.

HuStmpHrrr: I was trying to narrow it down to maybe just two instances and was experimenting with 
```coq
Fixpoint pointwise_rel (Ts : Tlist) (base : relation Prop) : binary_relation (predicate Ts) :=
  match Ts as Ts return binary_relation (predicate Ts) with
  | Tnil => base
  | Tcons T Ts' => @pointwise_relation T (predicate Ts') (pointwise_rel Ts' base)
  end.

#[export]
  Typeclasses Transparent pointwise_rel.

#[export]
  Instance subrelation_equivalence_iff Ts : subrelation (@predicate_equivalence Ts) (pointwise_rel Ts iff).
Proof.
  induction Ts; firstorder.
Qed.


#[export]
  Instance subrelation_equivalence_impl Ts : subrelation (@predicate_equivalence Ts) (pointwise_rel Ts Basics.impl).
Proof.
  induction Ts; firstorder.
Qed.
```

Coq doesn't seem to be smart enough but I though maybe it may simplify what it is now.

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -799,6 +771,34 @@ Proof.
...
+Qed.
+
+#[export]
+Instance subrelation_glu_exp_pred_equivalence_impl : subrelation glu_exp_pred_equivalence (@pointwise_relation ctx (predicate (Tcons exp (Tcons typ (Tcons domain Tnil)))) (@pointwise_relation exp (predicate (Tcons typ (Tcons domain Tnil))) (@pointwise_relation typ (predicate (Tcons domain Tnil)) (@pointwise_relation domain (predicate Tnil) Basics.impl)))).
+Proof.
+  intros ? ? ? ? ? ? ? ?. intuition.
+Qed.
+
+#[export]
+Typeclasses Transparent Basics.flip.
```
HuStmpHrrr, line 800: This is something I don't get. `iff_flip_impl_subrelation: iff ~> Basics.flip Basics.impl` is evidently in the hint database. I don't understand why resolution doesn't find it.

HuStmpHrrr, line 800:
```suggestion
Fixpoint pointwise_rel (Ts : Tlist) (base : relation Prop) : binary_relation (predicate Ts) :=
  match Ts as Ts return binary_relation (predicate Ts) with
  | Tnil => base
  | Tcons T Ts' => @pointwise_relation T (predicate Ts') (pointwise_rel Ts' base)
  end.

Lemma  subrelation_equivalence_iff Ts : subrelation (@predicate_equivalence Ts) (pointwise_rel Ts iff).
Proof.
  induction Ts; firstorder.
Qed.

#[export]
 Hint Extern 1 (subrelation (@predicate_equivalence ?Ts) _) => exact (subrelation_equivalence_iff Ts) : typeclass_instances.

#[export]
 Hint Extern 1 (subrelation iff Basics.impl) => exact iff_impl_subrelation : typeclass_instances.

#[export]
 Hint Extern 1 (subrelation iff (Basics.flip Basics.impl)) => exact iff_flip_impl_subrelation : typeclass_instances.
```

Ok. This is the best I've got. There is some problem in Coq when resolving iff to impl and flipped impl. That's why there are things that seem to not work. Not sure what's happening, but using extern does force it to work again.

HuStmpHrrr: The last two externs should probably be moved to LibTactics. Strange that this failure didn't happen before. Not sure if it's a bug or not.

Ailrun: Same here. I put this just because it works.

theories/Core/Soundness/LogicalRelation/Lemmas.v
Ailrun: Wait, does `pointwise_lifting iff l` also work for you if you put the transparenct hint on that? It didn't work for me last time I tried.

HuStmpHrrr: Check my suggested change below.

theories/Core/Soundness/LogicalRelation/Lemmas.v
Ailrun: You mean those externs work only with that custom fixpoint definition, not with pre-defined `pointwise_lifting`? If they work with `pointwise_lifting` it would be better to avoid the repeated definition.

HuStmpHrrr: ok, can we just prove the target instance just by applying the assumption? If so, then we don't even need this Fixpoint definition.

HuStmpHrrr: I am saying precisely this: `intro H; exact H`.

HuStmpHrrr: it's the same problem when resolving `iff ~> Basics.impl`. That's why you need all those extra `impl` instances. I believe if you set it transparent it will also work, but I think using the externs below is better.

HuStmpHrrr: ok. I think the extern is further improved.

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -799,6 +771,15 @@ Proof.
       * mauto 3.
 Qed.
 
+#[export]
+ Hint Extern 1 (subrelation (@predicate_equivalence ?Ts) _) => (let H := fresh "H" in intros ? ? H; exact H) : typeclass_instances.
```
HuStmpHrrr, line 775: Let's move these externs to LibTactics and call it a day. (I was not able to suggest changes in files that you didn't edit)

PR #151, opened 2024-08-01
Prove per escape lemma along gluing

Ailrun:
This will be used later for #136. More specifically, this one is a step towards subtyping escape lemma.

### DIFF
--- a/theories/Core/Soundness/EquivalenceLemmas.v
+++ b/theories/Core/Soundness/EquivalenceLemmas.v
@@ -1,0 +1,79 @@
+From Coq Require Import Morphisms Morphisms_Prop Morphisms_Relations Relation_Definitions RelationClasses.
+
+From Mcltt Require Import Base LibTactics.
+From Mcltt.Core.Syntactic Require Import Corollaries.
+From Mcltt.Core.Semantic Require Import Realizability.
+From Mcltt.Core.Soundness Require Import Realizability.
+From Mcltt.Core.Soundness Require Export LogicalRelation.
+Import Domain_Notations.
+
+Lemma glu_univ_elem_per_univ_typ_escape : forall {i a a' P P' El El' Γ A A'},
+    {{ Dom a ≈ a' ∈ per_univ i }} ->
+    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+    {{ DG a' ∈ glu_univ_elem i ↘ P' ↘ El' }} ->
+    {{ Γ ⊢ A ® P }} ->
+    {{ Γ ⊢ A' ® P' }} ->
+    {{ Γ ⊢ A ≈ A' : Type@i }}.
+Proof.
+  intros * [elem_rel Hper] Hglu Hglu' HA HA'.
+  gen A' A Γ. gen El' El P' P.
+  induction Hper using per_univ_elem_ind; intros; subst;
+    saturate_refl_for per_univ_elem;
+    invert_glu_univ_elem Hglu;
+    handle_functional_glu_univ_elem;
+    invert_glu_univ_elem Hglu';
+    handle_functional_glu_univ_elem;
+    try solve [simpl in *; mauto 4].
+  - destruct_by_head pi_glu_typ_pred.
+    rename M into M'. rename IT into IT'. rename OT into OT'.
+    rename x4 into IP'. rename x5 into IEl'. rename x6 into OP'. rename x7 into OEl'.
+    rename M0 into M. rename IT0 into IT. rename OT0 into OT.
+    rename x into IP. rename x0 into IEl. rename x1 into OP. rename x2 into OEl.
+    assert {{ Γ ⊢ IT ® IP }}.
+    {
+      assert {{ Γ ⊢ IT[Id] ® IP }} by mauto.
+      enough {{ Γ ⊢ IT[Id] ≈ IT : Type@i }} as <-; mauto.
+    }
+    assert {{ Γ ⊢ IT' ® IP' }}.
+    {
+      assert {{ Γ ⊢ IT'[Id] ® IP' }} by mauto.
+      enough {{ Γ ⊢ IT'[Id] ≈ IT' : Type@i }} as <-; mauto.
+    }
+    do 2 bulky_rewrite1.
+    assert {{ Γ ⊢ IT ≈ IT' : Type@i }} by mauto.
+    enough {{ Γ, IT ⊢ OT ≈ OT' : Type@i }} by mauto 3.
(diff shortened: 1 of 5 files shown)


theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -0,0 +1,78 @@
...
+    try solve [simpl in *; mauto 4].
+  - destruct_by_head pi_glu_typ_pred.
+    rename M into M'. rename IT into IT'. rename OT into OT'.
+    rename x4 into IP'. rename x5 into IEl'. rename x6 into OP'. rename x7 into OEl'.
+    rename M0 into M. rename IT0 into IT. rename OT0 into OT.
+    rename x into IP. rename x0 into IEl. rename x1 into OP. rename x2 into OEl.
+    assert {{ Γ ⊢ IT ® IP }}.
+    {
+      assert {{ Γ ⊢ IT[Id] ® IP }} by mauto.
+      enough {{ Γ ⊢ IT[Id] ≈ IT : Type@i }} as <-; mauto.
```
HuStmpHrrr, line 35: an autorewrite will eliminate `[Id]`

theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -0,0 +1,78 @@
...
+    assert {{ Γ ⊢ IT' ® IP' }}.
+    {
+      assert {{ Γ ⊢ IT'[Id] ® IP' }} by mauto.
+      enough {{ Γ ⊢ IT'[Id] ≈ IT' : Type@i }} as <-; mauto.
+    }
+    do 2 bulky_rewrite1.
+    assert {{ Γ ⊢ IT ≈ IT' : Type@i }} by mauto.
+    enough {{ Γ, IT ⊢ OT ≈ OT' : Type@i }} by mauto 3.
+    assert {{ Dom ⇑! A (length Γ) ≈ ⇑! A' (length Γ) ∈ in_rel }} as equiv_len_len' by (eapply per_bot_then_per_elem; mauto).
+    assert {{ Dom ⇑! A (length Γ) ≈ ⇑! A (length Γ) ∈ in_rel }} as equiv_len_len by (eapply per_bot_then_per_elem; mauto).
```
HuStmpHrrr, line 46: this and the one below can be derived by the first assert using saturate_refl

HuStmpHrrr: or saturate_refl_for

theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -0,0 +1,78 @@
...
+    destruct_rel_mod_eval.
+    handle_per_univ_elem_irrel.
+    rename a0 into b.
+    rename a' into b'.
+    assert {{ DG b ∈ glu_univ_elem i ↘ OP d{{{ ⇑! A (length Γ) }}} equiv_len_len ↘ OEl d{{{ ⇑! A (length Γ) }}} equiv_len_len }} by mauto.
+    assert {{ DG b' ∈ glu_univ_elem i ↘ OP' d{{{ ⇑! A' (length Γ) }}} equiv_len'_len' ↘ OEl' d{{{ ⇑! A' (length Γ) }}} equiv_len'_len' }} by mauto.
+    assert {{ Γ, IT ⊢ OT ® OP d{{{ ⇑! A (length Γ) }}} equiv_len_len }}.
+    {
+      assert {{ Γ, IT ⊢ #0 : IT[Wk] ® !(length Γ) ∈ glu_elem_bot i A }} by eauto using var_glu_elem_bot.
+      assert {{ Γ, IT ⊢ #0 : IT[Wk] ® ⇑! A (length Γ) ∈ IEl }} by (eapply realize_glu_elem_bot; mauto).
```
HuStmpHrrr, line 57: I think these two asserts are going to appear frequently enough to deserve its own lemma

theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -0,0 +1,78 @@
...
+      assert {{ Γ, IT ⊢ OT[Wk,,#0] ≈ OT : Type@i }} as <- by (bulky_rewrite1; mauto 3).
```
HuStmpHrrr, line 58: with your lemma below, won't autorewrite give you what you want already?

theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -0,0 +1,78 @@
...
+      assert {{ Γ, IT' ⊢ OT'[Wk,,#0] ® OP' d{{{ ⇑! A' (length Γ) }}} equiv_len'_len' }} by mauto.
+      assert {{ ⊢ Γ, IT' ≈ Γ, IT }} as <- by mauto.
+      assert {{ Γ, IT' ⊢ OT'[Wk,,#0] ≈ OT' : Type@i }} as <- by (bulky_rewrite1; mauto 3).
+      mauto.
+    }
+    mauto 3.
+  - match_by_head (per_bot b b') ltac:(fun H => specialize (H (length Γ)) as [V []]).
+    simpl in *.
+    destruct_conjs.
+    assert {{ Γ ⊢ A[Id] ≈ A : Type@i }} as <- by mauto 4.
```
HuStmpHrrr, line 74: autorewrite-able?

theories/Core/Soundness/EquivalenceLemmas.v
Ailrun: I manually derived them as I want names for those.

theories/Core/Soundness/EquivalenceLemmas.v
Ailrun: Unfortunately it does not. Putting `autorewrite with mcltt in *` or `autorewrite with mcltt in H<exact number>` instead of line 35 does not eliminate `[Id]`.

theories/Core/Soundness/EquivalenceLemmas.v
Ailrun: Note that this rewrites in the `<-` direction. No assumption of `OT[Wk,,#0]` is yet given. That will be latter proved by `mauto`. So it is basically the question of whether `assert` this or `assert` the goal with `OT[Wk,,#0]`.

theories/Core/Soundness/EquivalenceLemmas.v
Ailrun: No, at least not directly. The original goal is the equivalence between `A` and `A'`, and the easily obtainable assertions are equivalences between `A[Id]` (or `A'[Id]`) and `V`. To `autorewrite with mcltt in *` works, we first `assert` the equivalence between `A[Id]` and `V` *as well as* equivalence between `A'[Id]` and `V`. This requires 3 steps (two `assert`s and `autorewrite with mcltt in *`) just like here (three `assert`s). Also, we need to mention all names (`A`, `A'`, and `V`) even with such `autorewrite`.

PR #153, opened 2024-08-08
Prove subtyping case for soundness

Ailrun:
C.f. #136

~~`glu_univ_elem_per_subtyp_typ_if` requires a way to relate `OT'` (satisfying `{{ Γ, IT ⊢ OT' ® OP' d{{{ ⇑! a' (length Γ) }}} equiv_len'_len' }}`) and `OT''` (satisfying `{{ Δ ⊢ OT'' ® OP' c equiv_c }}`)
but I am not sure if it is even possible.~~

### DIFF
--- a/theories/Core/Soundness/Cumulativity.v
+++ b/theories/Core/Soundness/Cumulativity.v
@@ -7,6 +7,38 @@ From Mcltt.Core.Soundness Require Import Realizability.
 From Mcltt.Core.Soundness Require Export LogicalRelation.
 Import Domain_Notations.
 
+Corollary glu_univ_elem_cumu_ge : forall {i j a P El},
+    i <= j ->
+    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+    exists P' El', {{ DG a ∈ glu_univ_elem j ↘ P' ↘ El' }}.
+Proof.
+  intros.
+  assert {{ Dom a ≈ a ∈ per_univ i }} as [R] by mauto.
+  assert {{ DF a ≈ a ∈ per_univ_elem j ↘ R }} by mauto using per_univ_elem_cumu_ge.
+  mauto.
+Qed.
+
+#[export]
+Hint Resolve glu_univ_elem_cumu_ge : mcltt.
+
+Corollary glu_univ_elem_cumu_max_left : forall {i j a P El},
+    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+    exists P' El', {{ DG a ∈ glu_univ_elem (max i j) ↘ P' ↘ El' }}.
+Proof.
+  intros.
+  assert (i <= max i j) by lia.
+  mauto.
+Qed.
+
+Corollary glu_univ_elem_cumu_max_right : forall {i j a P El},
+    {{ DG a ∈ glu_univ_elem j ↘ P ↘ El }} ->
+    exists P' El', {{ DG a ∈ glu_univ_elem (max i j) ↘ P' ↘ El' }}.
+Proof.
+  intros.
+  assert (j <= max i j) by lia.
+  mauto.
+Qed.
+
 Section glu_univ_elem_cumulativity.
   #[local]
   Lemma glu_univ_elem_cumulativity_ge : forall {i j a P P' El El'},
--- a/theories/Core/Soundness/Cumulativity.v
+++ b/theories/Core/Soundness/Cumulativity.v
@@ -3,7 +3,7 @@ From Coq Require Import Morphisms Morphisms_Prop Morphisms_Relations Relation_De
 From Mcltt Require Import Base LibTactics.
 From Mcltt.Core.Syntactic Require Import Corollaries.
 From Mcltt.Core.Semantic Require Import Realizability.
-From Mcltt.Core.Soundness Require Import Realizability.
+From Mcltt.Core.Soundness Require Import EquivalenceLemmas Realizability.
 From Mcltt.Core.Soundness Require Export LogicalRelation.
 Import Domain_Notations.
 
@@ -14,7 +14,7 @@ Corollary glu_univ_elem_cumu_ge : forall {i j a P El},
 Proof.
   intros.
   assert {{ Dom a ≈ a ∈ per_univ i }} as [R] by mauto.
-  assert {{ DF a ≈ a ∈ per_univ_elem j ↘ R }} by mauto using per_univ_elem_cumu_ge.
+  assert {{ DF a ≈ a ∈ per_univ_elem j ↘ R }} by mauto.
   mauto.
 Qed.
 
@@ -48,7 +48,7 @@ Section glu_univ_elem_cumulativity.
       (forall Γ A, {{ Γ ⊢ A ® P }} -> {{ Γ ⊢ A ® P' }}) /\
         (forall Γ A M m, {{ Γ ⊢ M : A ® m ∈ El }} -> {{ Γ ⊢ M : A ® m ∈ El' }}) /\
         (forall Γ A M m, {{ Γ ⊢ A ® P }} -> {{ Γ ⊢ M : A ® m ∈ El' }} -> {{ Γ ⊢ M : A ® m ∈ El }}).
-  Proof.
+  Proof with mautosolve 4.
     simpl.
     intros * Hge Hglu Hglu'. gen El' P' j.
     induction Hglu using glu_univ_elem_ind; repeat split; intros;
@@ -65,8 +65,8 @@ Section glu_univ_elem_cumulativity.
       destruct_by_head pi_glu_typ_pred.
       econstructor; intros; mauto 4.
       + assert {{ Δ ⊢ IT[σ] ® IP }} by mauto.
-        assert (forall Γ A, {{ Γ ⊢ A ® IP }} -> {{ Γ ⊢ A ® IP' }}) by (eapply proj1; mauto).
-        mauto.
+        enough (forall Γ A, {{ Γ ⊢ A ® IP }} -> {{ Γ ⊢ A ® IP' }}) by mauto 4.
+        eapply proj1...
       + rename a0 into c.
         rename equiv_a into equiv_c.
         match_by_head per_univ_elem ltac:(fun H => directed invert_per_univ_elem H).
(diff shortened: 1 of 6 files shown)


theories/Core/Soundness/SubtypingLemmas.v
```diff
@@ -0,0 +1,164 @@
...
+    assert (exists OT'', {{ Δ ⊢ OT'' ® OP' c equiv_c }}) as [OT''] by mauto.
+    assert {{ Γ, IT ⊢ OT' ® glu_typ_top i b' }} as [] by mauto 4.
+    assert {{ Δ ⊢ OT'' ® glu_typ_top i b'0 }} as [] by mauto 4.
+    assert {{ DG b'0 ∈ glu_univ_elem i ↘ OP' c equiv_c ↘ OEl' c equiv_c }} by mauto.
+    enough {{ Δ ⊢ OT'[σ,,m] ≈ OT'' : Type@i }} as -> by mauto.
+    (* assert {{ Δ ⊢ OT''[Id] ≈ OT'' : Type@i }} as <- by mauto. *)
+    (* assert {{ Δ ⊢ m : IT[σ] }} by mauto 3 using glu_univ_elem_trm_escape. *)
+    (* assert {{ Δ ⊢s σ : Γ }} by mauto. *)
+    (* assert {{ Δ ⊢s Wk∘(Id,,m) ≈ Id : Δ }} as <- by mauto 3. *)
+    (* assert {{ Γ, IT ⊢ OT[Wk,,#0] ≈ OT : Type@i }} as <- by (bulky_rewrite1; mauto 3). *)
```
Ailrun, line 163: Here, we basically need to show that `pi_glu_typ_pred i in_rel IP IEl OP' Γ {{{ Π IT OT' }}}` is true if `pi_glu_typ_pred i in_rel IP IEl OP Γ {{{ Π IT OT }}}` where `OT` is a subtype of `OT'`.

Ailrun:
@HuStmpHrrr if you have any idea to make the stuck proof work, that would be appreciated.

HuStmpHrrr:
Why do you need to come up with this type? I tend to believe you won't need the lemmas you are trying to prove here.

theories/Core/Soundness/SubtypingLemmas.v
```diff
@@ -0,0 +1,164 @@
...
+From Mcltt.Core.Soundness Require Export LogicalRelation EquivalenceLemmas.
+Import Domain_Notations.
+
+Lemma glu_univ_elem_per_subtyp_typ_escape : forall {i a a' P P' El El' Γ A A'},
+    {{ Sub a <: a' at i }} ->
+    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+    {{ DG a' ∈ glu_univ_elem i ↘ P' ↘ El' }} ->
+    {{ Γ ⊢ A ® P }} ->
+    {{ Γ ⊢ A' ® P' }} ->
+    {{ Γ ⊢ A ⊆ A' }}.
```
HuStmpHrrr, line 16: instead of the lemma below, I think you need another version of this lemma but for El. It should have the same premises, but the conclusion should go from El to El', and the related type from A to A'. This should be the lemma to apply in the subsumption case.

Ailrun:
As I mentioned here: https://github.com/Beluga-lang/McLTT/issues/136#issuecomment-2251843278
For subtyping cases, we need `El` to `El'`, which then requires `P` to `P'` (as `El`s depend on `P`).

HuStmpHrrr:
right. I am saying `glu_univ_elem_per_subtyp_typ_escape` is all you need. you can just move on to prove the lemma about El. you probably won't need this lemma you are stuck at.

Ailrun:
Now I think that might be the case. Let me check the actual subtyping case. If that's indeed the case, I will remove the stuck lemma and then the rest can be merged

HuStmpHrrr:
yeah, I think the subtyping judgment is directly given by the semantic judgment in completeness, so you directly have the fact that the evaluations of two types are also in a subtyping relation. I think you can take it from there. You don't need a lemma that comes up with a type.

Ailrun:
@HuStmpHrrr Unfortunately, we still need a lemma for `P`. In the actual case, we only have `Δ ⊢ A[σ] ® P`, but the new lemma you suggested from `Δ ⊢ M[σ] : A[σ] ® m ∈ El` to `Δ ⊢ M[σ] : A'[σ] ® m ∈ El'` also depends on `Δ ⊢ A'[σ] ® P'`. I guess we need something like the following:
```
Lemma glu_univ_elem_per_subtyp_typ_if : forall {i Γ σ ρ Sb A A' a a' P P' El El'},
    {{ EG Γ ∈ glu_ctx_env ↘ Sb }} ->
    {{ Δ ⊢s σ ® ρ ∈ Sb }} ->
    {{ ⟦ A ⟧ ρ ↘ a }} ->
    {{ ⟦ A' ⟧ ρ ↘ a' }} ->
    {{ Sub a <: a' at i }} ->
    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
    {{ DG a' ∈ glu_univ_elem i ↘ P' ↘ El' }} ->
    {{ Γ ⊢ A ® P }} ->
    {{ Γ ⊢ A' ® P' }}.
```

Ailrun:
Well, no... that one is not provable...

HuStmpHrrr:
I think all you need is to add two premises to the subtyping rule, which says `A` and `A'` have type `Type@i`. Then you will have the right IHs which give you what you want. I will probably change some proofs of those syntactic lemmas, but to a livable extent.

theories/Core/Soundness/SubtypingCases.v
```diff
@@ -0,0 +1,74 @@
...
+  exists i,
+  forall Δ σ ρ,
+    {{ Δ ⊢s σ ® ρ ∈ Sb }} ->
+    glu_rel_exp_sub i Δ M A σ ρ.
+Proof.
+  intros * ? [Sb'].
+  destruct_conjs.
+  eexists; intros.
+  (* TODO: handle functional glu_ctx_env here *)
+  assert (Sb <∙> Sb') by admit.
```
Ailrun, line 23: Here

theories/Core/Soundness/SubtypingCases.v
```diff
@@ -0,0 +1,74 @@
...
+             | __mark__ _ _ => fail 1
+             | _ => edestruct H; try eassumption; mark H
+             end
+         end; unmark_all.
+  simplify_evals.
+  match_by_head glu_univ_elem ltac:(fun H => directed invert_glu_univ_elem H).
+  handle_functional_glu_univ_elem.
+  simpl in *.
+  (* TODO: introduce a lemma for glu_ctx_env *)
+  assert {{ Dom ρ ≈ ρ ∈ env_relΓ }} by admit.
```
Ailrun, line 55: And here

Ailrun (commented):
I leave the following `admit`s as they require more development of `glu_ctx_env` lemmas

HuStmpHrrr:
sorry will review today. stay tuned.

theories/Core/Soundness/Cumulativity.v
```diff
@@ -3,10 +3,42 @@ From Coq Require Import Morphisms Morphisms_Prop Morphisms_Relations Relation_De
 From Mcltt Require Import Base LibTactics.
 From Mcltt.Core.Syntactic Require Import Corollaries.
 From Mcltt.Core.Semantic Require Import Realizability.
-From Mcltt.Core.Soundness Require Import Realizability.
+From Mcltt.Core.Soundness Require Import EquivalenceLemmas Realizability.
 From Mcltt.Core.Soundness Require Export LogicalRelation.
 Import Domain_Notations.
 
+Corollary glu_univ_elem_cumu_ge : forall {i j a P El},
```
HuStmpHrrr, line 10: is this sufficient? you don't need a relation between P and P'?

theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -7,6 +7,73 @@ From Mcltt.Core.Soundness Require Import Realizability.
...
+  assert {{ Γ ⊢ A[Id] ≈ V : Type@(max i j) }} by mauto 4 using lift_exp_eq_max_left.
+  assert {{ Γ ⊢ A'[Id] ≈ V : Type@j }} by mauto 4.
+  assert {{ Γ ⊢ A'[Id] ≈ V : Type@(max i j) }} by mauto 4 using lift_exp_eq_max_right.
+  autorewrite with mcltt in *...
+Qed.
+
+#[export]
+Hint Resolve glu_univ_elem_typ_escape : mcltt.
+
+Lemma glu_univ_elem_typ_escape_ge : forall {i j a P P' El El' Γ A A'},
```
HuStmpHrrr, line 35: I think there can be a variant that completely ignores the relation between i and j. this variant might just be easier to apply.

theories/Core/Soundness/EquivalenceLemmas.v
```diff
@@ -7,6 +7,73 @@ From Mcltt.Core.Soundness Require Import Realizability.
 From Mcltt.Core.Soundness Require Export LogicalRelation.
 Import Domain_Notations.
 
+Lemma glu_univ_elem_typ_escape : forall {i j a P P' El El' Γ A A'},
+    {{ DG a ∈ glu_univ_elem i ↘ P ↘ El }} ->
+    {{ DG a ∈ glu_univ_elem j ↘ P' ↘ El' }} ->
+    {{ Γ ⊢ A ® P }} ->
+    {{ Γ ⊢ A' ® P' }} ->
+    {{ Γ ⊢ A ≈ A' : Type@(max i j) }}.
```
HuStmpHrrr, line 15: should this be max? I think it can even take min. say if i is smaller, than can't we bring A' down to i?

HuStmpHrrr:
I am sorry, this week was very bad for me. this PR looks almost ready to me.

theories/Core/Soundness/Cumulativity.v
Ailrun: This is just a helper as this is useful for later uses.

theories/Core/Soundness/EquivalenceLemmas.v
Ailrun: I am not sure that would be possible for auniverse case, but this is anyway enough for subtyping cases.

Ailrun: Yes, I think this should be max. In universe case (where `a` is the `k`th univ), this `P G A` (and `P' G A'`) corresponds to `A` (respectively `A') is equivalent to `Type@k` in `Type@i` (`Type@j`). Note that `A` may contains something that typable only in `Type@i` (also `A'` for `Type@j`).

theories/Core/Soundness/EquivalenceLemmas.v
Ailrun: See the above comment

PR #155, opened 2024-08-15
Rearrange soundness proofs

Ailrun:
Move lemmas independent from realizability into `CoreLemmas.v` and others into `Lemmas.v`.
Also I prove more lemmas about `glu_ctx_env`.


HuStmpHrrr (approved):
I don't have much to say.

PR #157, opened 2024-08-20
Change system rules

Ailrun:
This PR updates system rules so that it provides assumptions required for the soundness more naturally.
C.f. #121, #136

### DIFF
--- a/theories/Core/Syntactic/System/Definitions.v
+++ b/theories/Core/Syntactic/System/Definitions.v
@@ -82,6 +82,15 @@ with wf_exp : ctx -> typ -> exp -> Prop :=
      {{ Γ ⊢ M[σ] : A[σ] }} )
 | wf_exp_subtyp :
   `( {{ Γ ⊢ M : A }} ->
+     (** We have this extra argument for soundness.
+         Note that we need to keep it asymmetric:
+         only [A'] is checked. If we check A as well,
+         we cannot even construct something like
+         [{{ Γ ⊢ Type@0[Wk] : Type@1 }}] with the current
+         rules. Under the symmetric rule, the example requires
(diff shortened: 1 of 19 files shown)


theories/Core/Syntactic/System/Definitions.v
```diff
@@ -346,60 +348,141 @@ Inductive wf_ctx_eq : ctx -> ctx -> Prop :=
...
+    {{ Γ ⊢s σ : Δ' }}.
+Proof.
+  intros.
+  econstructor; mauto.
+Qed.
+
+#[export]
+Hint Resolve wf_sub_subtyp' : mcltt.
+#[export]
+Remove Hints wf_sub_subtyp : mcltt.
```
Ailrun, line 433: Remove extra arguments to keep the existing proofs as much as possible.

theories/Core/Syntactic/System/Definitions.v
```diff
@@ -346,60 +348,141 @@ Inductive wf_ctx_eq : ctx -> ctx -> Prop :=
...
 
-Add Parametric Morphism i Γ : (wf_exp Γ)
-    with signature wf_exp_eq Γ {{{ Type@i }}} ==> eq ==> iff as wf_exp_morph.
```
Ailrun, line 382: We cannot have this instance before presupposition lemmas as `wf_exp_eq` does not directly imply `wf_subtyp` before those.

HuStmpHrrr: oh yeah? It becomes this complicated? due to which change?

Ailrun: As `wf_subtyp_refl` requires the well-formedness of RHS, but that's only obtainable out of `wf_exp_eq` after the presuppositions.

Issue #113, opened 2024-06-12
Finish soundness of NbE


Ailrun:
- [x] Universe cases
  - #158
- [x] Context cases
  - #158
- [x] Term structure cases
  - #158
- [x] Subtyping cases
  - #153
  - #157
- [x] Substitution cases
  - #158
- [x] Function cases
  - #158
  - #159
- [ ] Natural number cases
  - #158
  - #160
- [x] Fundamental theorems
  - #156
  - #157

PR #158, opened 2024-08-20
Prove all easy soundness cases

Ailrun:
C.f. #113


HuStmpHrrr:
ok. what are the difficult cases?

Ailrun:
Natrec and function cases

PR #163, opened 2024-08-29
Add Algorithmic Typing Rules

### LINKED ISSUE #162: Algorithmic Typing Judgements
We need them for a readable presentation of our type-checker implementation.

Ailrun:
Resolves #162.
Together with completeness&soundness.

### DIFF
--- a/theories/Algorithmic/Typing/Lemmas.v
+++ b/theories/Algorithmic/Typing/Lemmas.v
@@ -20,6 +20,9 @@ Proof.
   reflexivity.
 Qed.
 
+#[local]
+Hint Resolve idempotent_nbe_ty : mcltt.
+
 Lemma functional_alg_type_infer : forall {Γ A A' M},
     {{ Γ ⊢a M ⟹ A }} ->
     {{ Γ ⊢a M ⟹ A' }} ->
@@ -42,6 +45,17 @@ Proof.
     functional_nbe_rewrite_clear.
     reflexivity.
 Qed.
+
+#[local]
+Hint Resolve functional_alg_type_infer : mcltt.
+
+Ltac functional_alg_type_infer_rewrite_clear1 :=
+  let tactic_error o1 o2 := fail 3 "functional_alg_type_infer equality between" o1 "and" o2 "cannot be solved by mauto" in
+  match goal with
+  | H1 : {{ ~?Γ ⊢a ~?M ⟹ ~?A1 }}, H2 : {{ ~?Γ ⊢a ~?M ⟹ ~?A2 }} |- _ =>
+      clean replace A2 with A1 by first [solve [mauto 2 using functional_alg_type_infer] | tactic_error A2 A1]; clear H2
+  end.
+Ltac functional_alg_type_infer_rewrite_clear := repeat functional_alg_type_infer_rewrite_clear1.
 
 Lemma alg_type_sound :
   (forall {Γ A M}, {{ Γ ⊢a M ⟸ A }} -> {{ ⊢ Γ }} -> forall i, {{ Γ ⊢ A : Type@i }} -> {{ Γ ⊢ M : A }}) /\
@@ -105,7 +119,7 @@ Lemma alg_type_infer_normal : forall {Γ A A' M},
     {{ Γ ⊢a M ⟹ A }} ->
     nbe_ty Γ A A' ->
     A = A'.
-Proof.
+Proof with (f_equiv; mautosolve 4).
   intros * ? Hinfer Hnbe. gen A'.
   assert {{ Γ ⊢ M : A }} by mauto 3 using alg_type_infer_sound.
   induction Hinfer; intros;
@@ -113,12 +127,10 @@ Proof.
          dir_inversion_by_head eval_exp; subst;
          dir_inversion_by_head read_typ; subst;
          reflexivity).
-  - eapply idempotent_nbe_ty; revgoals; try eassumption.
-    assert {{ Γ ⊢ ℕ : Type@0 }} by mauto 3.
+  - assert {{ Γ ⊢ ℕ : Type@0 }} by mauto 3.
     assert {{ Γ ⊢ M : ℕ }} by mauto 3 using alg_type_check_sound.
     assert {{ ⊢ Γ, ℕ }} by mauto 3.
-    assert {{ Γ, ℕ ⊢ A : ~n{{{ Type@i }}} }} by mauto 3 using alg_type_infer_sound.
-    mauto 4.
+    assert {{ Γ, ℕ ⊢ A : ~n{{{ Type@i }}} }} by mauto 3 using alg_type_infer_sound...
   - assert {{ Γ ⊢ A : ~n{{{ Type@i }}} }} by mauto 3 using alg_type_infer_sound.
     assert {{ Γ ⊢ A ≈ C : Type@i }} by mauto 3 using soundness_ty'.
     assert {{ Γ, A ⊢ M : B }} by mauto 3 using alg_type_infer_sound.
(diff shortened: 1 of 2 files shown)


theories/Algorithmic/Typing/Definitions.v
```diff
@@ -0,0 +1,71 @@
...
+
+Generalizable All Variables.
+
+Inductive alg_type_check : ctx -> typ -> exp -> Prop :=
+| atc_ati :
+  `( {{ Γ ⊢a M ⟹ A }} ->
+     {{ Γ ⊢a A ⊆ B }} ->
+     {{ Γ ⊢a M ⟸ B }} )
+where "Γ '⊢a' M ⟸ A" := (alg_type_check Γ A M) (in custom judg) : type_scope
+with alg_type_infer : ctx -> typ -> exp -> Prop :=
```
HuStmpHrrr, line 21: I think you can be more specific here by saying infer returns a normal form

theories/Algorithmic/Typing/Lemmas.v
```diff
@@ -0,0 +1,283 @@
...
+  pose proof alg_type_sound; intuition.
+Qed.
+
+Lemma alg_type_infer_sound : forall {Γ A M},
+    {{ Γ ⊢a M ⟹ A }} -> {{ ⊢ Γ }} -> {{ Γ ⊢ M : A }}.
+Proof.
+  pose proof alg_type_sound; intuition.
+Qed.
+
+Lemma alg_type_infer_normal : forall {Γ A A' M},
```
HuStmpHrrr, line 102: yeah, if you mark the type above to nf then this should be an obvious proof using idempotency.

theories/Algorithmic/Typing/Lemmas.v
```diff
@@ -0,0 +1,283 @@
...
+    {{ Γ ⊢a M ⟸ A' }}.
+Proof.
+  intros * [] **.
+  assert {{ Γ0 ⊢ B ⊆ A' }} by mauto 3.
+  assert {{ Γ0 ⊢a B ⊆ A' }} by mauto 3 using alg_subtyping_complete.
+  mauto 3 using alg_subtyping_trans.
+Qed.
+
+Lemma alg_type_complete : forall {Γ A M},
+    (* There should be one more condition to rule out substitution cases *)
```
HuStmpHrrr, line 236: yeah you will need a recursive predicate that says M contains no a_sub

theories/Algorithmic/Typing/Lemmas.v
Ailrun: How? I mean, even then it knows something like `nbe_ty Γ ? A` to apply idempotency, not just `A : nf`.

HuStmpHrrr: hmmm i see, but this lemma effectively also say that the inference mode gives *the* normal type, right? I somehow feel that the soundness lemma above can be strengthened to express that, though I don't have an idea about how now.

theories/Algorithmic/Typing/Lemmas.v
```diff
@@ -0,0 +1,302 @@
...
+    assert {{ Γ ⊢ B'[Id,,N] ⊆ B[Id,,N] }} by mauto 3.
+    assert {{ Γ ⊢ W ⊆ B[Id,,N] }} by (transitivity {{{ B'[Id,,N] }}}; mauto 3).
+    econstructor; mauto 4 using alg_type_check_conv, alg_subtyping_complete.
+  - assert (exists i, {{ Γ ⊢ A : Type@i }}) as [i] by mauto 3.
+    assert (exists W, nbe_ty Γ A W /\ {{ Γ ⊢ A ≈ W : Type@i }}) as [W []] by (eapply soundness_ty; mauto 3).
+    econstructor; mauto 4 using alg_subtyping_complete.
+Qed.
+
+(* There might be a better file for this. *)
+Functional Scheme elaborate_fun_ind := Induction for elaborate Sort Prop.
```
HuStmpHrrr, line 290: the same file as the elaborator?

theories/Core/Semantic/PER/Definitions.v
```diff
@@ -222,6 +222,10 @@ Section Per_univ_elem_ind_def.
         (fun _ _ _ _ _ => case_ne i)
         R a b H.
 
+  Search "per_univ_elem".
```
HuStmpHrrr, line 225: leftover

theories/Algorithmic/Typing/Lemmas.v
```diff
@@ -0,0 +1,302 @@
...
+    {{ Γ ⊢ A ≈ A' : Type@i }} ->
+    {{ Γ ⊢a M ⟸ A' }}.
+Proof.
+  intros * [] **.
+  assert {{ Γ0 ⊢ B ⊆ A' }} by mauto 3.
+  assert {{ Γ0 ⊢a B ⊆ A' }} by mauto 3 using alg_subtyping_complete.
+  mauto 3 using alg_subtyping_trans.
+Qed.
+
+Lemma alg_type_complete : forall {Γ A M},
```
HuStmpHrrr, line 238: I guess this lemma should also have an unfolded version, where the conclusion gives the infer direction, especially for A = some universe. This lemma will be applied when checking the user input type to be a valid type.

theories/Algorithmic/Typing/Definitions.v
```diff
@@ -0,0 +1,105 @@
...
+#[export]
+Hint Constructors alg_type_check alg_type_infer (* alg_univ_infer *) : mcltt.
+
+Scheme alg_type_check_mut_ind := Induction for alg_type_check Sort Prop
+with alg_type_infer_mut_ind := Induction for alg_type_infer Sort Prop.
+Combined Scheme alg_type_mut_ind from
+  alg_type_check_mut_ind,
+  alg_type_infer_mut_ind.
+
+Inductive user_exp : exp -> Prop :=
```
HuStmpHrrr, line 73: I think even this definition should go to the same file as the elaborator.

theories/Algorithmic/Typing/Definitions.v
```diff
@@ -0,0 +1,105 @@
...
+     {{ Γ ⊢a N ⟸ A }} ->
+     nbe_ty Γ {{{ B[Id,,N] }}} C ->
+     {{ Γ ⊢a M N ⟹ C }} )
+| ati_vlookup :
+  `( {{ #x : A ∈ Γ }} ->
+     nbe_ty Γ A B ->
+     {{ Γ ⊢a #x ⟹ B }} )
+where "Γ '⊢a' M ⟹ A" := (alg_type_infer Γ A M) (in custom judg) : type_scope.
+
+(* Variant alg_univ_infer : ctx -> nat -> typ -> Prop := *)
```
HuStmpHrrr, line 58: we will need this operation, but I don't suppose it should return anything. The `nat` can probably be removed.

theories/Algorithmic/Typing/Lemmas.v
Ailrun: You mean combining this with `alg_type_check_type_implies_alg_type_infer_type`? Sure, it's not hard.

theories/Algorithmic/Typing/Definitions.v
Ailrun: Why do we need this?

Ailrun: Just a FYI, the type checker is already implemented without `alg_univ_infer`.

HuStmpHrrr: well, where do you get a well-formed type to check against? you get it from the user right? so typechecking a file requires two calls to the type checker. one checks the well-formedness of user's input type, and the other checks the user's input term has this input type.

Ailrun: You can infer one. So, full type-checking steps are
- Construct WF of empty context
- Check the user-given type by Inferring its type.
- Check the user-given term against the empty context and the checked user-given type

PR #164, opened 2024-08-30
Implement type checker

Ailrun:
The extracted code for `type_check`/`type_infer` (under `ExtrOcamlBasic` and `ExtrOcamlNatBigInt`) is
```OCaml
(** val type_check : exp list -> exp -> exp -> bool **)

let rec type_check g a m =
  match type_infer g m with
  | Some a0 -> subtyping_impl g (nf_to_exp a0) a
  | None -> false

(** val type_infer : exp list -> exp -> nf option **)

and type_infer g = function
| A_zero -> Some Nf_nat
| A_succ e -> if type_check g A_nat e then Some Nf_nat else None
| A_natrec (e, e0, e1, e2) ->
  if type_check g A_nat e2
  then (match type_infer (A_nat :: g) e with
        | Some a ->
          (match a with
           | Nf_typ _ ->
             if type_check g (A_sub (e, (A_extend (A_id, A_zero)))) e0
             then if type_check (e :: (A_nat :: g)) (A_sub (e, (A_extend ((A_compose (A_weaken, A_weaken)), (A_succ (A_var
                       (Big_int_Z.succ_big_int Big_int_Z.zero_big_int))))))) e1
                  then Some (nbe_ty_impl g (A_sub (e, (A_extend (A_id, e2)))))
                  else None
             else None
           | _ -> None)
        | None -> None)
  else None
| A_nat -> Some (Nf_typ Big_int_Z.zero_big_int)
| A_typ n -> Some (Nf_typ (Big_int_Z.succ_big_int n))
| A_var n -> (match lookup g n with
              | Some a -> Some (nbe_ty_impl g a)
              | None -> None)
| A_fn (e, e0) ->
  (match type_infer g e with
   | Some a ->
     (match a with
      | Nf_typ _ -> (match type_infer (e :: g) e0 with
                     | Some a0 -> Some (Nf_pi ((nbe_ty_impl g e), a0))
                     | None -> None)
      | _ -> None)
   | None -> None)
| A_app (e, e0) ->
  (match type_infer g e with
   | Some s ->
     (match s with
      | Nf_pi (n, n0) ->
        if type_check g (nf_to_exp n) e0 then Some (nbe_ty_impl g (A_sub ((nf_to_exp n0), (A_extend (A_id, e0))))) else None
      | _ -> None)
   | None -> None)
| A_pi (e, e0) ->
  (match type_infer g e with
   | Some a ->
     (match a with
      | Nf_typ n ->
        (match type_infer (e :: g) e0 with
         | Some a0 -> (match a0 with
                       | Nf_typ n0 -> Some (Nf_typ (max n n0))
                       | _ -> None)
         | None -> None)
      | _ -> None)
   | None -> None)
| A_sub (_, _) -> None
```
and for `type_check_closed`,
```OCaml
(** val type_check_closed : exp -> exp -> bool **)

let type_check_closed a m =
  match type_infer [] a with
  | Some a0 -> (match a0 with
                | Nf_typ _ -> type_check [] a m
                | _ -> false)
  | None -> false
```

### DIFF
--- a/theories/Extraction/PseudoMonadic.v
+++ b/theories/Extraction/PseudoMonadic.v
@@ -1,0 +1,49 @@
+From Coq Require Extraction.
+
+(* We cannot use class based generalization for
+   the following definitions as Coq does not support
+   polymorphism across [Prop] and [Set] *)
+
+Definition sumbool_failable_bind {A B} (ab : {A} + {B}) {C D : Prop} (fail : B -> D) (next : A -> {C} + {D}) :=
+  match ab with
+  | left a => next a
+  | right b => right (fail b)
+  end.
+Transparent sumbool_failable_bind.
+Arguments sumbool_failable_bind /.
+
+Notation "'let*b' a ':=' ab 'while' fail 'in' next" := (sumbool_failable_bind ab fail (fun a => next)) (at level 0, a pattern, next at level 10, right associativity, only parsing).
+Notation "'pureb' a" := (left a) (at level 0, only parsing).
+
+Definition sumor_failable_bind {A B} (ab : A + {B}) {C} {D : Prop} (fail : B -> D) (next : A -> C + {D}) :=
+  match ab with
+  | inleft a => next a
+  | inright b => inright (fail b)
+  end.
+Transparent sumor_failable_bind.
+Arguments sumor_failable_bind /.
+
+Notation "'let*o' a ':=' ab 'while' fail 'in' next" := (sumor_failable_bind ab fail (fun a => next)) (at level 0, a pattern, next at level 10, right associativity, only parsing).
+Notation "'pureo' a" := (inleft a) (at level 0, only parsing).
+
+Definition sumbool_sumor_failable_bind {A B} (ab : {A} + {B}) {C} {D : Prop} (fail : B -> D) (next : A -> C + {D}) :=
+  match ab with
+  | left a => next a
+  | right b => inright (fail b)
+  end.
+Transparent sumbool_sumor_failable_bind.
+Arguments sumbool_sumor_failable_bind /.
+
+Notation "'let*b->o' a ':=' ab 'while' fail 'in' next" := (sumbool_sumor_failable_bind ab fail (fun a => next)) (at level 0, a pattern, next at level 10, right associativity, only parsing).
+
+Definition sumor_sumbool_failable_bind {A B} (ab : A + {B}) {C} {D : Prop} (fail : B -> D) (next : A -> {C} + {D}) :=
+  match ab with
+  | inleft a => next a
+  | inright b => right (fail b)
+  end.
+Transparent sumor_sumbool_failable_bind.
+Arguments sumor_sumbool_failable_bind /.
+
+Notation "'let*o->b' a ':=' ab 'while' fail 'in' next" := (sumor_sumbool_failable_bind ab fail (fun a => next)) (at level 0, a pattern, next at level 10, right associativity, only parsing).
+
+Extraction Inline sumbool_failable_bind sumor_failable_bind sumbool_sumor_failable_bind sumor_sumbool_failable_bind.
--- a/theories/Extraction/Subtyping.v
+++ b/theories/Extraction/Subtyping.v
@@ -1,13 +1,13 @@
 From Mcltt Require Import Base LibTactics.
 From Mcltt.Algorithmic Require Import Subtyping.Definitions.
 From Mcltt.Core Require Import NbE.
-From Mcltt.Extraction Require Import Evaluation NbE.
+From Mcltt.Extraction Require Import Evaluation NbE PseudoMonadic.
 From Equations Require Import Equations.
 Import Domain_Notations.
 
-
 #[local]
-  Ltac subtyping_tac :=
+Ltac subtyping_tac :=
+  intros;
   lazymatch goal with
   | |- {{ ⊢anf ~_ ⊆ ~_ }} =>
       subst;
@@ -23,22 +23,19 @@ Import Domain_Notations.
 
 #[tactic="subtyping_tac"]
 Equations subtyping_nf_impl A B : { {{ ⊢anf A ⊆ B }} } + {~ {{ ⊢anf A ⊆ B }} } :=
-| n{{{ Type@i }}}, n{{{ Type@j }}} with Compare_dec.le_lt_dec i j => {
-  | left _ => left _;
-  | right _ => right _
-  }
-| n{{{ Π A B }}}, n{{{ Π A' B' }}} with nf_eq_dec A A' => {
-  | left _ with subtyping_nf_impl B B' => {
-    | left _ => left _
-    | right _ => right _
-    }
-  | right _ => right _
-  }
+| n{{{ Type@i }}}, n{{{ Type@j }}} =>
+    let*b _ := Compare_dec.le_lt_dec i j while _ in
+    pureb _
+| n{{{ Π A B }}}, n{{{ Π A' B' }}} =>
+    let*b _ := nf_eq_dec A A' while _ in
+    let*b _ := subtyping_nf_impl B B' while _ in
+    pureb _
+(* Pseudo-monadic syntax for the next catch-all branch
+   generates some unsolved obligations *)
 | A, B with nf_eq_dec A B => {
   | left _ => left _
   | right _ => right _
   }.
-
 Theorem subtyping_nf_impl_complete : forall A B,
     {{ ⊢anf A ⊆ B }} ->
     exists H, subtyping_nf_impl A B = left H.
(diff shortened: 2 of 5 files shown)


theories/Extraction/TypeCheck.v
```diff
@@ -1,104 +1,392 @@
...
+  #[local]
+  Ltac impl_obl_tac := unfold not; intros; mauto 3 using user_exp_to_type_infer_order, type_check_order, type_infer_order.
+
+  #[tactic="impl_obl_tac",derive(equations=no,eliminator=no)]
+  Equations type_check_closed A (HA : user_exp A) M (HM : user_exp M) : { {{ ⋅ ⊢ M : A }} } + { ~ {{ ⋅ ⊢ M : A }} } :=
+  | A, HA, M, HM =>
+      let*o->b (exist _ UA _) := type_infer {{{ ⋅ }}} _ A _ while _ in
+      let*o->b (exist _ i _) :=  get_level_of_type_nf UA while _ in
+      let*b _ := type_check {{{ ⋅ }}} A _ M _ while _ in
+      pureb _
```
Ailrun, line 366: This is the entrypoint of the type checker. We need to elaborate `A : cst` and `M : cst` into `A' : typ` and `M' : exp`, and the elaboration should give `user_exp A'` and `user_exp M'` as well. From there, we can just call this function.

theories/Extraction/TypeCheck.v
```diff
@@ -1,104 +1,390 @@
...
+  | n{{{ Type@i }}} => pureo (exist _ i _)
+  | _               => inright _
+  .
+
+  Equations get_subterms_of_pi_nf (A : nf) : { '(B, C) | A = n{{{ Π B C }}} } + { forall B C, A <> n{{{ Π B C }}} } :=
+  | n{{{ Π B C }}} => pureo (exist _ (B, C) _)
+  | _              => inright _
+  .
+
+  Extraction Inline get_level_of_type_nf get_subterms_of_pi_nf.
```
Ailrun, line 58: Further simplification for type checkers

theories/Extraction/PseudoMonadic.v
```diff
@@ -0,0 +1,49 @@
...
+
+Definition sumbool_failable_bind {A B} (ab : {A} + {B}) {C D : Prop} (fail : B -> D) (next : A -> {C} + {D}) :=
+  match ab with
+  | left a => next a
+  | right b => right (fail b)
+  end.
+Transparent sumbool_failable_bind.
+Arguments sumbool_failable_bind /.
+
+Notation "'let*b' a ':=' ab 'while' fail 'in' next" := (sumbool_failable_bind ab fail (fun a => next)) (at level 0, a pattern, next at level 10, right associativity, only parsing).
```
HuStmpHrrr, line 15: what is letb and leto?

theories/Extraction/Subtyping.v
```diff
@@ -23,38 +23,33 @@ Import Domain_Notations.
...
-  }
-| n{{{ Π A B }}}, n{{{ Π A' B' }}} with nf_eq_dec A A' => {
-  | left _ with subtyping_nf_impl B B' => {
-    | left _ => left _
-    | right _ => right _
-    }
-  | right _ => right _
-  }
+| n{{{ Type@i }}}, n{{{ Type@j }}} =>
+    let*b _ := Compare_dec.le_lt_dec i j while _ in
```
HuStmpHrrr, line 27: what underscore becomes here?

theories/Extraction/PseudoMonadic.v
Ailrun: `let*b` is for `sumbool` (`b`ool) and `let*o` is for `sumor` (`o`r)

theories/Extraction/Subtyping.v
Ailrun: The first one (for `let*b _ :=`) is the `left _` case, and the second (`while _`) handles `right` case mapping.

theories/Extraction/TypeCheck.v
```diff
@@ -1,104 +1,390 @@
...
+
+  Inductive type_check_order : exp -> Prop :=
```
HuStmpHrrr, line 60: why do we even need this order? looks completely structural to me.

theories/Extraction/TypeCheck.v
```diff
@@ -1,104 +1,390 @@
...
+
+  Lemma user_exp_to_type_check_order : forall M,
+      user_exp M ->
+      type_check_order M.
+  Proof.
+    induction 1; progressive_inversion; mauto 3.
+  Qed.
+
+  Lemma user_exp_to_type_infer_order : forall M,
+      user_exp M ->
```
HuStmpHrrr, line 85: both lemmas together would allow you to just use `user_exp` to reject a_sub, if that's all you need.

theories/Extraction/TypeCheck.v
```diff
@@ -1,104 +1,390 @@
...
+  with type_infer G (HG : {{ ⊢ G }}) M (H : type_infer_order M) : { A : nf | {{ G ⊢a M ⟹ A }} /\ (exists i, {{ G ⊢a A ⟹ Type@i }}) } + { forall A, ~ {{ G ⊢a M ⟹ A }} } by struct H :=
+  | G, HG, M, H with M => {
+    | {{{ Type@j }}} =>
+        pureo (exist _ n{{{ Type@(S j) }}} _)
+    | {{{ ℕ }}} => 
+        pureo (exist _ n{{{ Type@0 }}} _)
+    | {{{ zero }}} =>
+        pureo (exist _ n{{{ ℕ }}} _)
+    | {{{ succ M' }}} =>
+        let*b->o _ := type_check G {{{ ℕ }}} _ M' _ while _ in
```
HuStmpHrrr, line 181: all I see is the `while` part is always unused. Is it always identity? If so, can we just simplify this and just define these lets to be more standard monadic do?

theories/Extraction/TypeCheck.v
HuStmpHrrr: what the extraction looks like now after you inline both definitions?

theories/Extraction/TypeCheck.v
Ailrun: Does Coq work with this kind of "redirection"? I mean, `type_check` basically calls `type_infer` while keeping the argument as-is.

theories/Extraction/TypeCheck.v
Ailrun: No, it is just a placeholder for a proof that we don't care in terms.

theories/Extraction/TypeCheck.v
Ailrun: Yeah, `struct M` does not work.

HuStmpHrrr: I think so, check itself is not recursive. it's just a helper function in some sense. I know fixpoint would probably have no problem. Equations shouldn't have problem too.

theories/Extraction/TypeCheck.v
Ailrun: I updated the original post

HuStmpHrrr:
```
     (match a with
      | Nf_pi (n, n0) ->
          let a0 = (n, n0) in
```
looks unnecessary. how is it generated?

Ailrun:
@HuStmpHrrr
It seems like this gives that.
```
Equations get_subterms_of_pi_nf (A : nf) : { '(B, C) | A = n{{{ Π B C }}} } + { forall B C, A <> n{{{ Π B C }}} } :=
  | n{{{ Π B C }}} => pureo (exist _ (B, C) _)
  | _              => inright _
```

theories/Extraction/TypeCheck.v
Ailrun: Unfortunately, no. `struct M` or skipping the annotation completely does not work for `type_check`.

HuStmpHrrr: Ok, I guess Agda is probably smarter in this part in checking termination

Ailrun: Yeah it is much smarter in this sense.

Ailrun:
What do you mean by curry it?

HuStmpHrrr:
instead of `{ '(B, C) | A = n{{{ Π B C }}} }`, try `{ B & { C | A = n{{{ Π B C }}} } }`

Ailrun:
@HuStmpHrrr With that, it has
```
(match a with
      | Nf_pi (n, n0) ->
        let a0 = ExistT (n, n0) in
        let ExistT (a1, s) = a0 in
```
so I guess it doesn't help in terms of the extracted code. However, it gives a better in-coq implementation, so I pushed it.

HuStmpHrrr:
> @HuStmpHrrr With that, it has
> ```
> (match a with
>       | Nf_pi (n, n0) ->
>         let a0 = ExistT (n, n0) in
>         let ExistT (a1, s) = a0 in
> ```
> so I guess it doesn't help in terms of the extracted code. However, it gives a better in-coq implementation, so I pushed it.

Wtf. This still sucks. I really wish something better can be done here.

Ailrun:
I don't see an easy way to resolve that unless we just directly inline the function even in Coq.

HuStmpHrrr:
if you are fine with it I think it's worth it to get rid of the glitch like this.

Ailrun:
You mean inlining the function? I don't think that's very optimal in terms of development, but the function is used once anyway... Ok, let's inline that.

HuStmpHrrr:
> You mean inlining the function? I don't think that's very optimal in terms of development, but the function is used once anyway... Ok, let's inline that.

Yeah, I think that's fine for this. I'm surprised the extraction doesn't take care of it.

Ailrun:
Updated.

PR #170, opened 2024-09-05
Add a more "proper" entrypoint with fuller parser

### LINKED ISSUE #110: Add surface language construct for natrec
We don't have one yet!

### LINKED ISSUE #172: Provide a proper entrypoint

Ailrun:
Resolves #110
Resolves #172

This one uses Coq 8.19.2 to avoid a conflict between Equations and Menhir.

To run CI for this, we need to update the Docker image first.


examples/nary.mcl
```diff
@@ -0,0 +1,49 @@
...
+  (
+    fun (a : Nat)
+        (b : Nat)
+      -> rec a return y -> Nat
+         | zero => b
+         | succ m, r => succ r
+         end
+  )
+)
+: Nat
```
Ailrun, line 49: We can encode this kind of "nary" example.

examples/nary.mcl
```diff
@@ -0,0 +1,49 @@
+(* main body *)
+(
+  fun (Nary : forall (n : Nat) -> Type@0)
```
HuStmpHrrr, line 3: what's the return type for this function? no need to specify it?

HuStmpHrrr: oh! this whole this is a single definition!

examples/nary.mcl
```diff
@@ -0,0 +1,49 @@
...
+      (appNary : forall (n : Nat) (f : Nary (succ n)) (arg : Nat) -> Nary n)
```
HuStmpHrrr, line 4: I would swap the order of this input and the next, just to line up the order below in the recursion.

theories/Core/Semantic/PER/Lemmas.v
```diff
@@ -535,8 +535,8 @@ Qed.
 Instance per_elem_PER {i R a b} `(H : per_univ_elem i R a b) : PER R.
 Proof.
   split.
-  - eauto using (per_elem_sym _ _ _ _ _ _ H).
-  - eauto using (per_elem_trans _ _ _ _ _ _ _ H).
+  - pose proof (fun m m' => per_elem_sym _ _ _ _ m m' H). eauto.
```
HuStmpHrrr, line 538: what's happening with these holes now?

theories/Entrypoint.v
```diff
@@ -0,0 +1,66 @@
...
+.
+
+Definition inspect {A} (x : A) : { y | x = y } := exist _ x eq_refl.
+Extraction Inline inspect.
+
+#[local]
+Ltac impl_obl_tac := try eassumption.
+#[tactic="impl_obl_tac"]
+Equations main (log_fuel : nat) (buf : buffer) : main_result :=
+| log_fuel, buf with Parser.prog log_fuel buf => {
```
HuStmpHrrr, line 34: is this fuel mandatory from the parser? maybe `50` is too small?

theories/Core/Semantic/PER/Lemmas.v
Ailrun: @HuStmpHrrr Coq 8.20 does not accept a term as a hint. So we need to provide a local assumption for that.

theories/Entrypoint.v
Ailrun: We can hard-code `50` there. I just wanted to specify that as later as possible.

examples/nary.mcl
Ailrun: Yeah, we don't have top-level definitions yet.

theories/Entrypoint.v
HuStmpHrrr: I mean let's set a large enough number, like a few million. there is no need to be too conservative i guess.

theories/Core/Semantic/PER/Lemmas.v
HuStmpHrrr: that's backwards

theories/Entrypoint.v
Ailrun: It's logarithmic version of fuel, so 2^50 should be liberative enough I believe

PR #171, opened 2024-09-05
Update Coq

Ailrun:
Coq 8.17.1 (or Coq-Equations supporting that version) has a compatibility issue with Coq-Menhirlib. When one tries to import two modules using each of Coq-Equations and Coq-Menhirlib, Coq initially gives an `"Anomaly: bad vernac extend:"` error. A funny thing is, it works after that, so for example
```coq
From Mcltt.Extraction Require Import TypeCheck.
Fail From Mcltt.Frontend Require Import Parser.
From Mcltt.Frontend Require Import Parser.
```
works. However, this seems very hacky and I have no idea how stable this would be, so it might be better to update Coq.

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,4 +1,4 @@
-FROM coqorg/coq:8.17.1-ocaml-4.14.1-flambda
+FROM coqorg/coq:8.20-ocaml-4.14-flambda
 
 ENV NCPUS=4
 ENV OPAMYES="true"
--- a/README.md
+++ b/README.md
@@ -47,18 +47,18 @@ implementation.
 
 ## Dependencies
 
-* OCaml 4.14.1
-* Menhir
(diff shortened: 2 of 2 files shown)


.github/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM coqorg/coq:8.17.1-ocaml-4.14.1-flambda
```
Ailrun, line 1: This docker image is outdated anyway. In fact, they, for some reason, do not support OCaml 4.14 for Coq 8.17 anymore.

Ailrun:
Or maybe we need to wait for a few days and jump to Coq 8.20

HuStmpHrrr:
> Or maybe it might be better to wait for a few days and jump directly to Coq 8.20

Ok then.

Issue #19, opened 2023-03-26
Redo makefile


HuStmpHrrr:
https://github.com/Beluga-lang/McLTT/blob/main/Makefile

Ailrun:
This is done afaik.

PR #179, opened 2024-09-09
update build process

HuStmpHrrr:
c.f. #19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,1 +1,14 @@
-# TODO: redo this makefile properly
+EXECUTABLE:=_build/default/driver/mcltt.exe
+
+.PHONY: all clean
+
+all:
+	@cd theories; make
+	@dune build
+
+clean:
+	@cd theories; make clean
+	@dune clean
+	@echo "Cleaning finished."
+
+${EXECUTABLE}: all
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,10 @@
-EXECUTABLE:=_build/default/driver/mcltt.exe
-
 .PHONY: all clean
 
 all:
-	@cd theories; make
+	@make -C theories
 	@dune build
 
 clean:
-	@cd theories; make clean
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1 +1,14 @@
-# TODO: redo this makefile properly
+EXECUTABLE:=_build/default/driver/mcltt.exe
+
+.PHONY: all clean
+
+all:
+	@cd theories; make
```
Ailrun, line 6: You can use `make -C theories`

Makefile
```diff
@@ -1 +1,14 @@
...
+	@dune build
+
+clean:
+	@cd theories; make clean
```
Ailrun, line 10: Likewise

Makefile
```diff
@@ -1 +1,14 @@
...
+	@dune clean
+	@echo "Cleaning finished."
+
+${EXECUTABLE}: all
```
Ailrun, line 14: Do we need this target? It seems up-side-down in the sense that `$(EXECUTABLE)` is a part of "build all".

driver/Mcltt.ml
```diff
@@ -0,0 +1,11 @@
...
+
+let () =
+  if Array.length Sys.argv <> 2 then
+    begin
+      Printf.fprintf stderr "incorrect input: %s <input-file>\n" Sys.argv.(0);
+      exit 1
+    end;
+  let filename = Sys.argv.(1) in
+  main ~default_fp:filename ();
+  exit 0
```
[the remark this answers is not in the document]
HuStmpHrrr: yeah no need

Makefile
HuStmpHrrr: done

Makefile
HuStmpHrrr: yeah probably not

Ailrun (approved):

Issue #145, opened 2024-07-29
Use CI to generate a reasonable homepage


HuStmpHrrr:
We can wait. Just a reminder that we need this.

The Coq code should also have a web version that is browsable. I think there is something called tactic animation or something similar. We should have that.

Ailrun:
I believe this is done?

PR #183, opened 2024-09-10
implement a homepage

HuStmpHrrr:
c.f. #145


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
HuStmpHrrr, line 1: this bot is needed to publish github page. to be done

PR #191, opened 2024-09-10
[WIP] working on ocaml doc

HuStmpHrrr:
It doesn't work as expected; not sure how to fix it


HuStmpHrrr:
`dune build @doc` doesn't do what one would expect.

Ailrun:
Let's see how https://github.com/ocaml/dune/issues/10921 goes.

HuStmpHrrr:
I think I will modify this PR a bit and merge it. The changes to dune files are alright so we can keep them. We probably don't want to generate the doc for ocaml for now.

PR #192, opened 2024-09-11
Remove warnings in extraction impls

Ailrun:


theories/Extraction/Evaluation.v
```diff
@@ -108,13 +108,13 @@ Qed.
...
 #[local]
-  Ltac impl_obl_tac :=
+Ltac impl_obl_tac :=
   repeat impl_obl_tac1; try econstructor; eauto.
 
 Derive Signature for eval_exp_order eval_natrec_order eval_app_order eval_sub_order.
 
-#[tactic="impl_obl_tac"]
-  Equations eval_exp_impl m p (H : eval_exp_order m p) : { d | eval_exp m p d } by struct H :=
+#[tactic="impl_obl_tac",derive(equations=no,eliminator=no)]
```
Ailrun, line 116: Note that we don't need any details about impl anyway. `eval_exp` is functional, so if it ever exists `d` is fixed. All we care about `eval_exp_impl` is its termination, which is given.

theories/Extraction/Evaluation.v
```diff
@@ -173,102 +173,55 @@ Derive Signature for eval_exp_order eval_natrec_order eval_app_order eval_sub_or
...
+          eexists; reflexivity
+      end
+  end.
 
 Lemma eval_exp_impl_complete : forall M p m,
     {{ ⟦ M ⟧ p ↘ m }} ->
     exists H H', eval_exp_impl M p H = exist _ m H'.
 Proof.
-  repeat unshelve mauto.
+  intros; functional_eval_complete.
```
HuStmpHrrr, line 205: we don't even need induction? how come?

Ailrun: destruction on `eval_exp_impl M p H` is enough; once it gives `exists _ m' H'`,  we can use functionality to get `m = m'`.

theories/Extraction/Evaluation.v
HuStmpHrrr: yeah I think the messages are about equation proofs.

theories/Extraction/Evaluation.v
HuStmpHrrr: i see

PR #193, opened 2024-09-11
recover timed CI

HuStmpHrrr:


Makefile
```diff
@@ -1,9 +1,13 @@
...
 
 all:
 	@$(MAKE) -C theories
 	@dune build
 
+pretty-timed:
+	@$(MAKE) pretty-timed -C theories
+	@dune build
+
 coqdoc: all
```
Ailrun, line 11: This might cause 2 extractions

PR #201, opened 2024-09-15
Optimizing proofs

Ailrun:
Let's see how much this saves.


Ailrun:
About 30secs? Not bad.

PR #234, opened 2024-10-01
Simple let-bindings

Antoine-something:
Change "->" to "." in recursion
Add let bindings that are parsed into function application


HuStmpHrrr:
There is a conflict in the PR.

HuStmpHrrr:
I also think that I might be worth it to change the complex examples to this let binding form.

driver/Test.ml
```diff
@@ -48,3 +48,11 @@ let%test "pi missing colon" = parse "pi (x Nat).x" = None
 let%test "ignore whitespace" =
   parse "fun (x  \n                                     : Type 4).x"
   = Some (Coq_fn (x, Coq_typ 4, Coq_var x))
+
+
+(* Tests for let-binders *)
+let%test "let" =
+  parse "let x : Type 2 = Type 1 in x" = Some (Coq_app (Coq_fn (x, Coq_typ 2, Coq_var x), Coq_typ 1)
```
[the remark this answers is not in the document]
Ailrun: I guess the test hasn't been run?

Ailrun:
It would be better to add that `examples/simple_let.mcl` as a test in `Test.ml` like
```OCaml
let%expect_test "simple_let.mcl works" =
  let _ = main_of_example "simple_let.mcl" in
  [%expect {| |}]
```
After that, you can run `dune test` to check that the mismatch happens only for that test (as well as that program output is correct), and `dune promote` will write the output back to the test file.

Ailrun:
Some examples should also be fixed

HuStmpHrrr:
please also remember to fix the pretty printer.

theories/Frontend/Parser.vy
```diff
@@ -34,16 +34,19 @@ let obj :=
...
   | ~ = app_obj; <>
 
-  | REC; escr = obj; RETURN; mx = VAR; "->"; em = obj;
+  | REC; escr = obj; RETURN; mx = VAR; "."; em = obj;
     "|"; ZERO; "=>"; ez = obj;
     "|"; SUCC; sx = VAR; ","; sr = VAR; "=>"; ms = obj;
     END; { Cst.natrec escr (snd mx) em ez (snd sx) (snd sr) ms }
   | SUCC; ~ = obj; { Cst.succ obj }
 
+  | LET; p = param; ":="; arg_obj = obj; IN; fun_obj = obj; { Cst.app (Cst.fn (fst p) (snd p) fun_obj) arg_obj }
```
Ailrun, line 43: It would be more useful to extend this into multiple concurrent definitions so that instead of generating
`(fun x -> (fun y -> x + y) 6) 5` out of
```
let x = 5 in
let y = 6 in
x + y
```
we can generate `(fun x y -> x + y) 5 6`. This is useful for the current "datatype" examples (which provide datatype interface in the latter way).

HuStmpHrrr: may be just change the grammar to capture a list of `let x = ? in` and then convert it to the desired ast? introducing new syntax sounds unnecessary.

Ailrun: i don't think `(fun x y -> x + y) 5 6` is always desirable. Sometimes we want to define `y` using `x`, and either of that requires a new syntax or some syntactical overhead.

Antoine-something: There is no real difference between `let (x := 5) (y := 6) in x + y` and `let (x := 5) and (y := 6) in x + y`, so I don't see why we would need new syntax. If we have dependencies like `let (x := 5) (y := x + 1) in x + y`, the problem is that after parsing to `(fun x y -> x + y) 5 (x + 1)` has `x` out of scope in `y`, which is not solved just with new syntax. 

Right now, my parser can handle the lists of definitions, but I don't quite see how to deal with the dependencies, appart from using two `let`s like `let (x := 5) in let (y := x + 1) in x + y`, which is exactly what we would like to avoid.  It seems we would need the parser to perform substitutions in order to handle such dependencies, with is also probably not desirable.

HuStmpHrrr: yeah I agree. the current solution is temporary and will be replaced by genuine local bindings with delta reduction. I think it's better to focus on the outstanding problems and get this PR merged asap.

HuStmpHrrr:
a small thing is you obviously have two github accounts. you probably want to choose one.

theories/Frontend/Parser.vy
Ailrun: @Antoine-something This PR doesn't support `let (x := 5) (y := 6) in x + y`, right? I don't care about detailed syntax. I'm just talking about difference between
```
let (x : Type@0) = Nat in
let (y : x) = 6 in
(fun (a : x) -> a) y
```
and
```
let ((x : Type@0) = Nat)
  ((y : x) = 6) in
(fun (a : x) -> a) y
```
Depending on translation, only one of them will be accepted.

But anyway, let's merge this PR first and then deal with this later. Even though this is a temporary solution, I think we will use this for a while, so it would be better for this solution to deal with the readability of our examples.

HuStmpHrrr: But even we encode the latter example in your proposed way, it still won't type check, no?

Ailrun: It type checks if I understand correctly. The function part (say `f`) is fine, and `f Nat` gives `Nat -> Nat` so `f Nat 6` is also fine.

Antoine-something: Yeah that's right, this PR does not support `let (x := 5) (y := 6) in x + y`, you would have to use two `let`.  I meant that in my local repo, I have support for it so that it expands to `(fn x y -> x + y) 5 6`.  I haven't updated the PR yet because I still have to modify the pretty printer.

Also, I rewrote the nary and vector examples to use the `let` bindings, and they parse exactly like the non-`let` versions.

Antoine-something: I'm looking at the pretty parser now.  If I understand correctly, it only prints after parsing, which means that all the `let`s have been turned into function applications.  Is that right?  If so, I don't think there is anything to do with it.

Ailrun: It might be better to print let syntax if there is a function immediately applied to an argument.

HuStmpHrrr: I wouldn't go down to that route. Let's not do more than necessary. I was referring to the changes from -> to . Let's keep a faithful pretty printer.

PR #253, opened 2024-10-16
remove redundant check

### LINKED ISSUE #252: redundant comparison in readback and PER
https://github.com/Beluga-lang/McLTT/blob/9ed1a85ae63e1f1fb5cbd297af509bce18bd800b/theories/Core/Semantic/Readback/Definitions.v#L23

the inner check on `ℕ` is unnecessary. Since readback is type directed, this inner check necessarily passes because we only pass in well-typed terms for normalization. 

In fact, 

https://github.com/Beluga-lang/McLTT/blob/9ed1a85ae63e1f1fb5cbd297af509bce18bd800b/theories/Core/Semantic/PER/Definitions.v#L69

even here we don't even need to care about `ℕ` because we are relating two values of type Nat, so we can safely just say for all domain types, and we know for sure they are going to be `ℕ`.

HuStmpHrrr:
Closes #252

### DIFF
--- a/theories/Core/Soundness/NatCases.v
+++ b/theories/Core/Soundness/NatCases.v
@@ -402,7 +402,7 @@ Proof.
   assert {{ ⊢ Δ, ℕ }} by mauto 3.
   assert {{ Δ, ℕ ⊢ A[q σ] : Type@i }} by mauto 3.
   assert {{ ⊢ Δ, ℕ, A[q σ] }} by mauto 2.
-  unshelve assert {{ Δ ⊢ M : ℕ }} by mauto 3; [constructor |].
+  assert {{ Δ ⊢ M : ℕ }} by mautosolve 3.
   assert {{ Δ ⊢ ℕ : Type@0 }} by mauto 3.
   assert {{ Δ ⊢ ℕ[σ] ≈ ℕ : Type@0 }} by mauto 3.
   assert {{ Δ ⊢ M : ℕ[σ] }} by mauto 3.
--- a/theories/Core/Soundness/NatCases.v
+++ b/theories/Core/Soundness/NatCases.v
@@ -352,7 +352,7 @@ Qed.
 #[local]
 Hint Resolve cons_glu_sub_pred_q_nat_helper : mcltt.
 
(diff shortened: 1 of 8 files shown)


HuStmpHrrr:
current extraction:

```ocaml
(** val read_nf_impl : nat -> domain_nf -> nf **)

let rec read_nf_impl s = function
| D_dom (d0, d1) ->
  (match d0 with
   | D_nat ->
     (match d1 with
      | D_zero -> Nf_zero
      | D_succ d2 -> Nf_succ (read_nf_impl s (D_dom (D_nat, d2)))
      | D_neut (_, d2) -> Nf_neut (read_ne_impl s d2)
      | _ -> assert false (* absurd case *))
| ...
```
the type annotation is safe to ignore for the neutral case.

theories/Core/Soundness/NatCases.v
```diff
@@ -402,7 +402,7 @@ Proof.
   assert {{ ⊢ Δ, ℕ }} by mauto 3.
   assert {{ Δ, ℕ ⊢ A[q σ] : Type@i }} by mauto 3.
   assert {{ ⊢ Δ, ℕ, A[q σ] }} by mauto 2.
-  assert {{ Δ ⊢ M : ℕ }} by mauto 3.
+  unshelve assert {{ Δ ⊢ M : ℕ }} by mauto 3; [constructor |].
```
Ailrun, line 405: Using `mautosolve 3` instead of `mauto 3` is not enough?

Ailrun (approved):
Other than the above comment, LGTM.

HuStmpHrrr:
Not sure. Does it do u shelve? I don't even know when this unresolved goal is coming from.

Thanks,
Jason Hu
________________________________

Ailrun:
It solves shelved goals as well, so it should work if it is generated because of the universe of Nat.

HuStmpHrrr:
it's not really about universe levels. It asks for a `domain`. I am not sure which lemma it's coming from, but is definitely caused by my change. but in any case, `mautosolve` does handle it.

Issue #181, opened 2024-09-09
Support propositional equality


HuStmpHrrr:
We should add propositional equality. There are three terms:

1. `Eq A a b` which is the type;
2. `refl A a` is the only constructor;
3. `J` is the elimination principle; check ITT for an exact formulation.

We need the following things:

- [ ] Add syntactic rules
- [ ] Add cases for the domain model
- [ ] Extend evaluation and readback
- [ ] Fix the PER model
- [ ] Fix the completeness proof
- [ ] Fix the Kripke gluing model
- [ ] Fix the soundness proof

HuStmpHrrr:
We use `ext/prop-eq` for this extension, and then eventually merge it to `main`.

HuStmpHrrr:
As per 
![image](https://github.com/user-attachments/assets/450b2ea2-cfa0-4ad8-9c99-043387192750)
we do not need to change the subtyping rule at all for equality.

c.f. https://coq.inria.fr/doc/V8.18.0/refman/language/cic.html#subtyping-rules

also

```coq
Axiom x y : Set.

Definition foo : Type := x.

Axiom bar : @eq Set x y.

Fail Definition baz : @eq Type x y := bar.
```

HuStmpHrrr:
thinking about it twice, I think we cannot let subtyping to propagate in any inductive data types. consider just equality. If we allowed subtyping so that if `A <: B`, then `Eq A a b <: Eq B a b`, then we immediately lose canonicity of equality. because now `refl A a` can also be a constructor of `Eq B a a`, while otherwise, `refl B a` is the only possible constructor. Clearly, `refl A a` and `refl B a` cannot be related. we should really be careful about our intuition with the presence of subtyping.

Ailrun:
How about contravariant subtyping for pi types? It has a similar problem: `\lambda B 5` is of `\Pi B \BbbN`, but not directly of `\Pi A \BbbN` (for a subtype `A` of `B`)

HuStmpHrrr:
I tend to believe negative types are better off due to eta expansion. The same probably applies to products.

Ailrun:
So you mean any positive types should not allow subtyping? Hm, that might be the case...

Ailrun:
Soundness is not just "tedious". It requires quite a significant change in proofs. For one of **simpler** examples, we need to change `glu_univ_elem_typ_monotone` and `glu_univ_elem_exp_monotone` to be mutually proved.

PR #259, opened 2024-11-04
Working on soundness proof

HuStmpHrrr:
c.f. #181

### DIFF
--- a/theories/Core/Soundness/LogicalRelation/CoreLemmas.v
+++ b/theories/Core/Soundness/LogicalRelation/CoreLemmas.v
@@ -658,37 +658,87 @@ Ltac invert_glu_univ_elem H :=
    destruct H as [? [? [? [? [? [? [? [? []]]]]]]]])
   + basic_invert_glu_univ_elem H.
 
-Lemma glu_univ_elem_resp_per_univ : forall i a a' P El,
-    {{ Dom a ≈ a' ∈ per_univ i }} ->
+Lemma glu_nat_resp_per_nat : forall m n,
+    {{ Dom m ≈ n ∈ per_nat }} ->
+    forall Γ M,
+      glu_nat Γ M m ->
+      glu_nat Γ M n.
+Proof.
+  induction 1; intros; progressive_inversion; mauto.
+  econstructor.
+  - mauto using (PER_refl2 _ per_bot).
+  - intros.
+    specialize (H (length Δ)).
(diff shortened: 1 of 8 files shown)


HuStmpHrrr:
@Ailrun I realize I don't have much time because my days are occupied now. I should probably behave more like your shadow. Let me know what I should look into that does not require immediate unblock.

theories/Core/Soundness/LogicalRelation/CoreLemmas.v
```diff
@@ -656,6 +678,12 @@ Proof.
     handle_per_univ_elem_irrel.
     intuition.
   - reflexivity.
+  - mauto using (PER_refl2 _ R).
+  - mauto using (PER_refl2 _ R).
+  - split; intros []; econstructor; intuition.
```
HuStmpHrrr, line 683: my fix is until this point, where this lemma needs to be proved mutually with a case for transporting El using R generated by `per_univ_elem`.

theories/Core/Soundness/LogicalRelation/CoreLemmas.v
```diff
@@ -656,6 +678,12 @@ Proof.
...
+
+
+
   - apply neut_glu_typ_pred_morphism_glu_typ_pred_equivalence.
```
HuStmpHrrr, line 687: need similar counterparts for `eq` above.

theories/Core/Soundness/LogicalRelation/CoreLemmas.v
Ailrun: If I remember correctly it requires realizability, which makes thing more complex.

HuStmpHrrr: why? doesn't seem to have proved this lemma even?

Ailrun: If it's not the case, all good, but that's what I remember. Of course it could be a wrong memory.

HuStmpHrrr: yeah this needs no realizability.

HuStmpHrrr:
@Ailrun ok it's ready for now. let's do another PR for other fixes.

Ailrun (approved):
LGTM

PR #271, opened 2024-12-10
finished semantic well-formedness of equality type

HuStmpHrrr:
c.f. #181

### DIFF
--- a/theories/Core/Soundness/LogicalRelation/Lemmas.v
+++ b/theories/Core/Soundness/LogicalRelation/Lemmas.v
@@ -650,7 +650,7 @@ Proof.
     handle_functional_glu_univ_elem.
     trivial.
 Qed.
-xb
+
 Lemma glu_univ_elem_per_subtyp_trm_conv : forall {i j k a a' P P' El El' Γ A A' M m},
     {{ Sub a <: a' at i }} ->
     {{ DG a ∈ glu_univ_elem j ↘ P ↘ El }} ->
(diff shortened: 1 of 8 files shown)


theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -1227,12 +1227,13 @@ Proof.
   eapply glu_univ_elem_exp_conv; revgoals; mauto 3.
 Qed.
 
+#[global]
```
Ailrun, line 1230: Why global (not export)?

theories/Core/Soundness/LogicalRelation/Lemmas.v
```diff
@@ -650,7 +650,7 @@ Proof.
     handle_functional_glu_univ_elem.
     trivial.
 Qed.
-
+xb
```
Ailrun: typo here

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: confusion I guess

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: fixed

HuStmpHrrr:
ok I will pause here and do another PR for the last lemma.

theories/Core/Soundness/LogicalRelation/Lemmas.v
HuStmpHrrr: ok. I guess I was doubly confused. ltac doesn't export.

PR #274, opened 2024-12-24
Work on the final semantic rule for equality

HuStmpHrrr:


HuStmpHrrr:
need two things:
- [ ] An inversion helper for types in semantics that are being substituted (should be easy);
- [ ] Injectivity of Eq (shouldn't be hard).

theories/Core/Soundness/EqualityCases.v
```diff
@@ -93,6 +93,27 @@ Qed.
...
+    {{ Γ ⊩ N : Eq A M1 M2 }} ->
+    glu_rel_exp_resp_sub_env i Sb N {{{Eq A M1 M2}}}.
+Proof.
+  intros * ? HA HM1 HM2 HN.
+  assert {{ Γ ⊩ Eq A M1 M2 : Type@i }} by mauto.
+  eapply glu_rel_exp_clean_inversion2 in HN; eassumption.
+Qed.
+
+
+#[global]
```
Ailrun, line 110: As last time you mentioned, export is not for Ltac, but isn't export is the default for Ltac?

jiangsy:
Shall I close this as it's subsumed by #284 ?

PR #280, opened 2025-05-21
finish holes in completeness for equality

jiangsy:


theories/Core/Completeness/EqualityCases.v
```diff
@@ -184,29 +185,44 @@ Proof.
...
+  destruct_conjs.
+  match_by_head per_eq ltac:(fun H => destruct H).
+  destruct_by_head rel_typ.
+  destruct_by_head rel_exp.
+  rename ρ1 into ρσ.
+  rename ρ0 into ρ'σ.
+  simplify_evals.
+  - eexists; split. 2: {
+      (* is this two assertions true? *)
+      assert ({{ ⟦ BR ⟧ ρσ ↦ n ↘ m7 }}) by admit.
```
jiangsy, line 216: @Ailrun Is this true that evaluating `A` under related domains gives the same value? if so, are there any existing lemmas/proof schemes to prove this or for reference?

jiangsy: ah, I see. I probably should assert like `{{ Dom ρ1 ↦ m1 ≈ ρ0 ↦ m0 ∈ env_relΔA }}` using different values in the first place

jiangsy: Is this true that evaluating A under related domains gives the same value <- OK, this is not true I guess, only get related values. I kinda understand what you mean by "insane number of steps", though.

Ailrun:
Oh, really? If that's the case, it's a great news.

Ailrun:
Oh, BTW, after merging this PR, could you use a differently named branch for development? `ext/...` is meant to be a main branch for a specific extension, and they causes a webpage update for every push. Something like `feature/prop-eq-complete` or whatever else you prefer.

PR #281, opened 2025-05-23
fill holes in completeness

jiangsy:


jiangsy:
it seems we need to add `Γ, A, A[Wk], Eq A[Wk∘Wk] #1 #0 ⊢ B : Type@j` to `wf_exp_eq_eqrec_cong` because `Γ, A, A[Wk], Eq A[Wk∘Wk] #1 #0 ⊨ B : Type@j ` is helpful to the proof of completeness. I think it's fine to add it, right?

Ailrun:
You can obtain that at the beginning using symmetry instead of changing the syntactic judgement

jiangsy:
symmetry is valid for PER, but not, at least trivially, for the semantic judgment itself? SImilarly, `{{ Γ ⊨ M1 : A }}` is also included in the premise, though we already have `Γ ⊨ M1 ≈ M1' : A`

jiangsy:
the completeness is all done. some possible known issues: (1) usage of UIP (#282); (2) significantly longer type checking (9 mins alone for `Completeness/EqCases.v`, more than the total of all the remaining); (3) some non-robust tactics that are brittle to changes, like `apply H58`

theories/Core/Completeness/EqualityCases.v
```diff
@@ -128,6 +129,254 @@ Qed.
...
+  etransitivity; [eassumption |].
+  intros s.
+  match_by_head per_bot ltac:(fun H => specialize (H s) as [? []]).
+  eexists; split; [eassumption |].
+  dir_inversion_by_head read_ne; subst.
+  simplify_evals.
+  mauto.
+Qed.
+
+Lemma equiv_rel_PER : forall {A R R'},
```
HuStmpHrrr, line 358: move this somewhere else? I think we have a place for lemmas like this?

theories/Core/Semantic/PER/CoreTactics.v
```diff
@@ -6,16 +6,23 @@ From Mctt.Core Require Import Base.
 From Mctt.Core.Semantic Require Import PER.Definitions.
 Import Domain_Notations.
 
+Ltac lazy_destruct H H' :=
+  match type of H with
+  | forall _ _ _, exists _, _ => pose proof (H _ _ H'); deex_once
```
HuStmpHrrr, line 11: why 3? this is too specific?

PR #290, opened 2025-06-26
Add assumptions checking to CI

jiangsy:


jiangsy:
it's not used now. I didn't figure out the best way to integrate it yet. 

a simple solution could be execute the script before `make` and filter all the make outputs to check if all axiom names are expected

scripts/insert_assumption_checks.py
```diff
@@ -0,0 +1,54 @@
+# python insert_assumption_checks.py ../
```
Ailrun, line 1: Is this meant to be a `#!`?

scripts/check_assumptions.py
```diff
@@ -0,0 +1,61 @@
...
+
+    for line in lines:
+        stripped = line.strip()
+
+        # Match: <<<./path/to/file.v - lemma_name>>>
+        marker_match = re.match(r'^<<<(.+?) - (.*)>>>$', stripped)
+        if marker_match:
+            path = marker_match.group(1)
+            lemma = marker_match.group(2)
+            current_name = f"{lemma}"
```
HuStmpHrrr, line 26: what's the reason for this? isn't lemma a string?

scripts/check_assumptions.py
```diff
@@ -0,0 +1,61 @@
...
+            continue
+
+        axiom_name_match = re.match(r'^(.*) :$', stripped)
+        if axiom_name_match:
+            axiom_name = axiom_name_match.group(1)
+            if axiom_name in expected_axiom and current_name is not None:
```
HuStmpHrrr, line 32: why check `current_name`? it's clearly not `None`.

scripts/insert_assumption_checks.py
```diff
@@ -0,0 +1,53 @@
...
+
+        qed_pattern = re.compile(r'.*(Qed\.|Defined\.|Admitted\.)', re.MULTILINE)
+        qed_match = qed_pattern.search(content, pos=start_pos)
+
+        if qed_match:
+            insert_pos = qed_match.end()
+            marker = f"{rel_path} - {name}"
+            snippet = (
+                f'\n{indentation}Goal True. idtac "<<<{marker}>>>". Abort.\n'
+                f'{indentation}Print Assumptions {name}.'
```
HuStmpHrrr, line 27: is this how we print assumptions? why can't we just get the list of lemma names and print them all, instead of modifying all files?

jiangsy: Ah, I get your point. You are proposing let the CI generate a `***.v` file that contains all `Print`s? We could do that.

HuStmpHrrr:
> It shows `No unexpected axiom usage.` under the `Check axioms usage` group. The allowed axioms can be configured in the python script and are also printed out in CI.
> 
> ```
>   Warning: theorem glu_univ_elem_cumulativity_ge in Core/Soundness/LogicalRelation/Lemmas.v marked with #[local] is skipped.
>   Warning: theorem ctxsub_exp_helper in Core/Syntactic/CtxSub.v marked with #[local] is skipped.
>   Allowed Axioms are ['FunctionalExtensionality.functional_extensionality_dep'].
>   No unexpected axiom usage.
> ```
> 
> I cannot think of more to add now.

ok. it seems we also have two warnings that are easy to fix. is it ready to merge?

jiangsy:
the warnings are a bit irrelevant. it's output by the script to show that it does not add theorems marked with `#[local]` into the final check file.

it's ready to merge

HuStmpHrrr (approved):

Issue #296, opened 2025-08-16
Add a Sigma Type


jiangsy:

jiangsy:
IIUC, there are two ways to define the computation rules of sigma types (https://ncatlab.org/nlab/show/dependent+sum+type) by viewing it as either negative types or positive types.

1. have $\pi_1$ and $\pi_2$, given $c:=(a, b) : \Sigma A , B(a)$, then $\pi_1(c) \equiv a$ and  $\pi_2(c) \equiv b$, also $(\pi_1(c),(\pi_2(c)) \equiv c$

2. have $\texttt{let}\ (x,y) = c~\texttt{in}~d$, given $c:=(a, b) : \Sigma A , B(a)$, then $\texttt{let}(x,y)=c\ \texttt{in}\ d \equiv d[a/x, b/y]$ also $\texttt{let}\ (x,y) =\ c\ \texttt{in}\ d[(x,y)/z] \equiv d[c/z]$,

I would prefer the first one. do you have any comments on this?

jiangsy:
I am thinking about can we force the syntax of pairs to be `< M , N : B >` instead of `< M , N >` which again makes pairs inferrable, and their implementation independent of #298

Ailrun:
You mean we can add `B` that is open (by one variable)? That sounds like harder to describe on a paper.

HuStmpHrrr:
> I am thinking about can we force the syntax of pairs to be `< M , N : B >` instead of `< M , N >` which again makes pairs inferrable, and their implementation independent of [#298](https://github.com/Beluga-lang/McTT/issues/298)

I am for this change. this is natural and what I was thinking about initially as well. one thing that I think should be done is to define a weak notion of equalities between nf and ne which ignore certain fields, e.g. input types of lambdas, and in this case, the type of the second component of a pair.

PR #297, opened 2025-08-18
Add definitions for Sigma types

jiangsy:
#296

### DIFF
--- a/theories/Core/Semantic/Readback/Definitions.v
+++ b/theories/Core/Semantic/Readback/Definitions.v
@@ -30,12 +30,13 @@ Inductive read_nf : nat -> domain_nf -> nf -> Prop :=
      {{ Rnf ⇓ b m' in S s ↘ M }} ->
      {{ Rnf ⇓ (Π a ρ B) m in s ↘ λ A M }} )
 | read_nf_sigma : 
-  `( (** Normal form of snd type *)
+  `( (** Normal form of snd type, 
+         should be abstract over the first element *)
      {{ ⟦ B ⟧ ρ ↦ ⇑! a s ↘ b }} ->
      {{ π₁ m ↘ m' }} ->
      {{ π₂ m ↘ n }} ->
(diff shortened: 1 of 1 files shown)


Ailrun:
I changed the base branch

.github/workflows/ci_build.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
   build:
     name: Continuous Intergration
     runs-on: ubuntu-latest
+    needs: [wf]
```
Ailrun, line 28: This breaks the CI (as I disabled `wf` for now)

jiangsy: not really. but I think anyway it does not make much sense to run CI on unfinished codebase, so as long as the files are commented out in _CoqProject, we can skip the whole CI

jiangsy:
I think I can still go with the annotated syntax currently, i.e. `⟨ M , N : x . B ⟩` as it should be consistent with the current core theory. And we can remove all type annotations all together later.

Ailrun:
Sure, that makes work more concurrent.

theories/Core/Semantic/Evaluation/Definitions.v
```diff
@@ -77,6 +94,19 @@ with eval_app : domain -> domain -> domain -> Prop :=
   `( {{ ⟦ B ⟧ ρ ↦ n ↘ b }} ->
      {{ $| ⇑ (Π a ρ B) m & n |↘ ⇑ b (m (⇓ a n)) }} )
 where "'$|' m '&' n '|↘' r" := (eval_app m n r) (in custom judg)
+with fst_app : domain -> domain -> Prop :=
```
jiangsy, line 97: `fst_app` is not actually mutually inductively defined with other relations. It's a bit weird but I cannot tell anything wrong. any comments on this?

Ailrun: This seems right. It's just because the first projection does nothing to do with environment/context.

Issue #298, opened 2025-08-20
Separate core and surface syntax


jiangsy:
After introducing sigma types #297, we would have checkable-only (or non-inferrable) terms. To properly type-check of them, we need to add annotations (M : A) into syntax to allow programmers give annotations to arbitrary terms. However, we do not want to add this syntax to our type theory, thus, we need to level of syntax and build a equivlance-up-to-erasure relation between our core (declarative, type assignment) system and the algorithm (bidirectional).

jiangsy:
related #202

Ailrun:
The IR approach indeed suffers from a problem of correctness. Suppose that we have an expression `(M : A)` where `M` is a metavariable for IR and its internal `exp` form is `M'`.
Now, if we have `Γ ⊢ M' : B`, the only thing we can get out of that is `exists M'', {{ Γ ⊢a M'' ⟸ B }}` and the internal expression of IR `M''` is `M`. We lose the relationship between `M` and `M'` here.
In other words, this breaks completeness of our type-checking algorithm.

I guess we need to introduce annotation on the core.

Ailrun:
- Adding it to core does not resolve the issue
- One way to possibly resolve this is using `elab M' : A' ~ M : A` where `elab_:_~_:_` is a relation of type `ir -> ir -> exp -> typ -> Prop`. Although I am not sure whether this really works...

PR #299, opened 2025-08-21
[Annotation] Extract CST into a separate module

Ailrun:
As the first step of #298, this PR first extracts CST into a separate module (so that the addition of other syntactic layer can be done independently from the Core module hierarchy)


Ailrun:
Oops, I meant to squash it, but anyway.

PR #301, opened 2025-08-24
Experiment on contra-variant subtyping

jiangsy:


HuStmpHrrr:
fix the cr by commenting entries in _CoqProject, and then it can be merged.

jiangsy:
> https://github.com/Beluga-lang/McTT/tree/pr-some-contra This branch may also give you some idea

Thanks for the reminder. I have already ported most development in that branch.

Ailrun:
Could you explain why do we need that weak cong?

Ailrun:
Also some relevant discussion: https://github.com/Beluga-lang/McTT/issues/129

jiangsy:
> Could you explain why do we need that weak cong?

it's still an experiment. basically for "realizaibility" of subtyping, I think we want `⇑ a n` and `⇑ a' n'` where `a <: a'` (instead of `a ≈ a'`) to be "related" in some sense. but the readback of reflection with a sub/super value could (potentially) change the argument annotation. so we would like to prove a weaker lemma to show that except for these annotations, we get the same output.

Ailrun:
That sounds more like we need to remove the annotation on lambda after introducing type annotations on an intermediate representation

theories/Core/Semantic/WRealizability.v
```diff
@@ -0,0 +1,201 @@
...
+    dependent destruction H7.
+    invert_per_univ_elem H5.
+    invert_per_univ_elem H8.
+    destruct_all.
+    intros s.
+    assert {{ Dom ⇑! a'0 s ≈≈ ⇑! a'1 s ∈ in_rel }}. {
+      eapply H17; mauto 3; saturate_refl;
+      eapply per_subtyp_refl; eauto.
+    }
+    (* we cannot construct related values that belong to
```
jiangsy, line 80: @HuStmpHrrr

jiangsy: I think the problem is that the forall quantification reverse the expected direction of per_subtyp

HuStmpHrrr: I can infer the state by looking at it. it seems to work when I worked on the board. we will need to look at it together.

PR #306, opened 2025-09-01
Remove weakening from non-neutral cases

Ailrun:
Related: #303


Ailrun:
This alone, without any other aggressive optimization, gives 10% speed up in CI.

PR #309, opened 2025-09-02
Update the Algo for Sigma Types

jiangsy:


jiangsy:
I think there's a problem in for pair

```
Lemma alg_type_check_complete : forall {Γ A M},
    user_exp M ->
    {{ Γ ⊢ M : A }} ->
    {{ Γ ⊢a M ⟸ A }}.
```

For the pair case, `{{ Γ ⊢a ⟨ M; N : B ⟩ ⟸ Σ A B }}`, with IH we know `{{ Γ ⊢a M ⟸ A }}`, then we can only concluded `exists A', {{ Γ ⊢a M ⟹ A' }} /\ {{ Γ ⊢a A' ⊆ A }`. but with `Σ A' ... ⊆ Σ A B` will be rejected since our sigma subtyping requires  `A' ≈ A`.

Ailrun:
It seems like that's the case. We probably want to add the separate general annotation sooner than later...

jiangsy:
the algorithm is now good

Ailrun (approved):
LGTM

PR #310, opened 2025-09-03
Rework Sigma Types

jiangsy:


theories/Core/Completeness/PairCases.v
```diff
@@ -122,14 +122,15 @@ Qed.
 #[export]
 Hint Resolve rel_exp_sigma_sub : mctt.
 
-Lemma rel_exp_pair_cong : forall {i Γ A B B' M N M' N'},
+Lemma rel_exp_pair_cong : forall {i Γ A A' B B' M N M' N'},
     {{ Γ ⊨ A : Type@i }} ->
+    {{ Γ ⊨ A ≈ A' : Type@i }} ->
```
Ailrun, line 127: This one should replace the one before.

Ailrun (approved):
Other than the one above, LGTM

theories/Core/Completeness/PairCases.v
jiangsy: actually this one is not even used now, the previous one alone is enough

jiangsy: but both premises can be obtained from syntactic rules. do you think we should make the syntactic rules more precise in the first place?

Ailrun: Ah true. Ok, let's keep it this way and see if we can make it more succinct later.

PR #312, opened 2025-09-04
Implement sigma parser

Ailrun:
This PR also changes equality types' syntax so that the type of compared terms looks more like an implicit argument


jiangsy:
to mark here, UIP is re-introduced in paircases currently

https://github.com/Beluga-lang/McTT/actions/runs/17475331085/job/49633694069#step:7:879

jiangsy:
I'm fixing the UIP problem

Ailrun:
> btw, do you think we should fail the CI if unexpected axioms appear?
Yes, I think once we restore no-UIP status, we should also change CI so that it fails on UIP (and any other unexpected axioms).

jiangsy:
Sure, I'll fix CI in another PR. I think this one should be good to go after we do a manual check on CI results

jiangsy (approved):
LGTM

PR #313, opened 2025-09-04
Make the script error out when finding unexpected axioms

jiangsy:
1. capture errors from subprocess output and print them
2. signal an error if unexpected axioms appear


jiangsy:
it works as expected now

Issue #269, opened 2024-12-07
Include global definitions


HuStmpHrrr:
For McTT to be more realistic, we cannot avoid to include global definitions. Having global definitions also allows us to have custom data types and axioms. 

Adding global definitions should be simple: in addition to `Gamma`, the local context, we include a global context, say, `Sigma`. It is worth thinking about the structure of `Sigma`. I think it would be the best for it to have a tree form to take modules into account. Essentially, `Sigma` is a rose tree. Each tree is a map from string to a node. Each node could be either a leaf, which include a type and an optional body, or a subtree. In this way, we can implement modules quite seamlessly.

It is also easy to implement other features, like `abstract` in Agda, by just forgetting the body at all.

I can imagine the following necessary changes:

1. The elaborator should resolve names to local de Bruijn indices and global, fully quantified names.
2. We should include delta equivalence for global definitions that has a body.
3. The evaluation proceed must include now a global environment. I think global environments should have a simpler structure than local ones, because there is no need to consider substitutions. Details need to be pinned down.
4. The proof then needs to take global contexts and environments into account. I am not sure how PER models should react. We should try to understand it better.

Ailrun:
https://github.com/Beluga-lang/McTT/tree/ext/top-level branch is created for this

HuStmpHrrr:
Like I have outlined above, we should think about implementing modules as well. This shouldn't add more complication. A global name should be a non-empty list of strings that is a path to search modules organized as a tree. 

Local let binding should be implemented similarly.

PR #321, opened 2025-10-09
Adding global contexts to syntax

jiangsy:
related #269

### DIFF
--- a/theories/Core/Syntactic/Syntax.v
+++ b/theories/Core/Syntactic/Syntax.v
@@ -1,4 +1,5 @@
 From Coq Require Import List String.
+From stdpp Require Import gmap.
 
 From Mctt.Core Require Import Base.
 
@@ -24,6 +25,7 @@ Inductive obj : Set :=
           string -> obj ->                   (** x. Pf : M[x/x, x/y, refl A x/z] *)
           obj -> obj -> obj -> obj
 | var : string -> obj.
+(* TODO: do we need gvar for this? possibly yes *)
 End Cst.
 
 (** * Abstract Syntac Tree *)
@@ -51,6 +53,8 @@ Inductive exp : Set :=
 | a_eqrec : exp -> typ -> exp -> exp -> typ -> exp -> exp
 (** Variable *)
 | a_var : nat -> exp
+(* gvar for global var *)
+| a_gvar : string -> exp
 (** Substitution Application *)
 | a_sub : exp -> sub -> exp
 with sub : Set :=
@@ -61,6 +65,18 @@ with sub : Set :=
 where "'typ'" := exp.
 
 Notation ctx := (list exp).
+
+(* We store the name, defintion (optional, none represents axiom), and type of the defintion *)
+
+Notation gctx := (list (string * option exp * typ)).
+
+Fixpoint gctx_dom (Δ : gctx) : gset string :=
+  match Δ with
+  | nil => ∅
+  | cons (x, _, _) Δ' => {[x]} ∪ gctx_dom Δ'
+  end.
+
+(** ** Substitution *)
 
 Fixpoint nat_to_exp n : exp :=
   match n with
@@ -104,6 +120,7 @@ with ne : Set :=
 | ne_fst : ne -> ne
 | ne_snd : ne -> ne
 | ne_var : nat -> ne
+| ne_gvar : string -> ne
 | ne_eqrec : nf -> nf -> nf -> nf -> nf -> ne -> ne
 .
 
@@ -128,6 +145,7 @@ with ne_to_exp (M : ne) : exp :=
   | ne_fst M => a_fst (ne_to_exp M)
   | ne_snd M => a_snd (ne_to_exp M)
   | ne_var x => a_var x
+  | ne_gvar x => a_gvar x
   | ne_eqrec A B BR M M' N =>
       a_eqrec (nf_to_exp A) (nf_to_exp B) (nf_to_exp BR) (nf_to_exp M) (nf_to_exp M') (ne_to_exp N)
   end
@@ -142,7 +160,8 @@ with ne_eq_dec : forall (M M' : ne),
     ({M = M'} + {M <> M'})%type.
 Proof.
   all: intros; decide equality;
-    apply PeanoNat.Nat.eq_dec.
+    first [ apply PeanoNat.Nat.eq_dec
+           | apply string_dec ].
 Defined.
 
 Definition q σ := a_extend (a_compose σ a_weaken) (a_var 0).
@@ -185,6 +204,7 @@ Module Syntax_Notations.
   Notation "'eqrec' N 'as' 'Eq' A M1 M2 'return' B | 'refl' -> BR 'end'" := (a_eqrec A B BR M1 M2 N) (in custom exp at level 0, A custom exp at level 30, B custom exp at level 60, BR custom exp at level 60, M1 custom exp at level 35, M2 custom exp at level 40, N custom exp at level 60) : mctt_scope.
 
   Notation "'#' n" := (a_var n) (in custom exp at level 0, n constr at level 0, format "'#' n") : mctt_scope.
+  Notation "'`#' x" := (a_gvar x) (in custom exp at level 0, x constr at level 0, format "'`#' x") : mctt_scope.
 
   Notation "'Id'" := a_id (in custom exp at level 0) : mctt_scope.
   Notation "'Wk'" := a_weaken (in custom exp at level 0) : mctt_scope.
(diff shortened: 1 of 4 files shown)


Ailrun:
If we have `nbe G A W W`, then what would be the problem? It is reflexivity, isn't it?

HuStmpHrrr:
what would be your step rule for global contexts?

jiangsy:
> what would be your step rule for global contexts?

do you mean the delta reduction rule?

HuStmpHrrr:
setting body to be an option is a representation of axioms. this is just a very simple change for a boost of practical use.

jiangsy:
I think we are concurrently arguing about 4 choices (update me if I miss some points), so let me make them a bit clear. 

1. store exp/nf in the global context
2. use strings/de bruijn levels as names
3. use flat (list)/rose tree structure global context
4. allow optional definitions

Ailrun:
> I am referring to the typing rule. this requires the cons case of the typing rule for global contexts, which is defined mutually with existing rules, to refer to NbE, which is only defined much later. I am not sure how you are able to write down the typing rules. I wouldn't suggest to perform any extra operation, if simplicity is desired.

Declaratively, we have no issues as we can simply assume that globals are already all normalized. Note that NbE itself (not its properties) does not depend on typing. In algorithm, we can add one to global context only after normalizing a term.

HuStmpHrrr:
> > the main focus of this change should be what is the right form of explicit substitutions. with global names as strings, this is easy: it's a syntactic model for mappings from strings to terms, satisfying a permutation rule.
> 
> I don't think strings or de-bruijn levels make any difference in this case? because we only subst a name with its def in the delta-equality rule, and retrive the def from the global context is the same whether it's string or a natural number.
> 
> also, are you suggesting we only put closed terms into the global context? in this case, I think a rose-tree global context does not give any extra benefit over a flat one.
> 
> for storing exp or nf, either is OK for me. w.r.t to the optional def, I think you are right. but does having axioms, i.e. blocked terms as nf (in the non-empty global context), cause new complexties? if not, adding that would be easy.

it matters, when you think of axioms and re-printing. how do you recover the name, if you wiped it off? this case happens if you consider axioms. as I said, this is almost for free: you just allow an optional definition bound to a global name. an axiom is a global definition without a body.

jiangsy:
@Ailrun will you catch up with Brigitte offline this week? we had a quick online meeting yesterday but it would be helpful if you also sync with her

HuStmpHrrr:
This will be my last response to this PR, so this is the best argument I can make.

Global names as de Bruijn indices/levels blocks all of following three features:

1. axioms,
2. modules,
3. true inductive types as I define and care about, which might not be of your interest, in this given context. In this case, inductive types are composed of two things: constructors as axioms, and an eliminator that reacts on these axioms.

Implementing any of these features necessitates a total rework of what will get in this PR. This includes reconsideration of explicit global substitutions (I assume it is the case, as I don't think anyone of you is interested in turning to substitutions as operations, as I remembered), related rules, and related semantic rules, which cover everything this PR is about. As an initial implementation, I can imagine, a tree map from strings to whatever bound to global names suffices. It can clearly be extended for any use above and more. Coq has its tree map library, doing it is as free as using lists. There is no benefit to do trial and failure, if problems have been foreseen. 

Writing Coq code is time consuming. Building something that lasts and extensible is of great importance. No one will go back and prettify Coq code. So please design something that stays. 

At last, I am still asking if anyone has written down any wf rules for global contexts. If you write them down, you will probably understand why my preferences go that way. In contrary, I am reading Coq's typing rule, which is completely expected and reasonable to me: https://rocq-prover.org/doc/V8.18.0/refman/language/cic.html

Read `W-Global-Assum` and `W-Global-Def`.

Frankly, most design discussions occurring here don't make much sense to me, as there is a quite obvious way to do it. There hasn't been many controversies around these corners. It's likely the same feeling for reviewers. Get ready to provide good reasons.

theories/Core/Syntactic/System/Definitions.v
```diff
@@ -22,464 +25,488 @@ Inductive ctx_lookup : nat -> typ -> ctx -> Prop :=
   | there : `({{ #n : A ∈ Γ }} -> {{ #(S n) : A[Wk] ∈ Γ, B }})
 where "'#' x : A ∈ Γ" := (ctx_lookup x A Γ) (in custom judg) : type_scope.
 
-Inductive wf_ctx : ctx -> Prop :=
-| wf_ctx_empty : {{ ⊢ ⋅ }}
+Inductive gctx_lookup : string -> exp -> typ -> gctx -> Prop :=
+  | ghere : `({{ #x := M :: A ∈ Γ, x := M :: A }})
+  | gthere : `(x <> y -> {{ #y := M :: A ∈ Γ }} -> {{ #y := M :: A ∈ Γ, x := N :: B }})
```
Ailrun: This also needs to deal with axioms. I think `exp` should be `exp option`?

jiangsy:
To summarize, we are currently using `list (string * option exp * typ)` for global contexts, as our potential target - metacoq also uses lists instead of trees.

jiangsy:
the following constructors, in various inductive relations, are added: `wf_gctx_empty`, `wf_gctx_extend`, `wf_gctx_extend_ax`, `wf_gvlookup`, `wf_exp_eq_delta`, `wf_exp_eq_gvar_ax_refl`

I don't think we need an inductive relation for wf_gctx_eq

@Ailrun can you double check?

HuStmpHrrr:
I have the following points to raise:

1. why `wf_ctx_sub` needs to take two global contexts? this is not useful, as global contexts in practice only grows in one direction, e.g. only weakening applies. also, during type-checking, you never change a global context. a second global context seems an overshoot to me.
2. what's the thought on global substitutions? don't we need it? if not, what's the reason behind?
3. if explicit global substitutions are needed, then there needs to be an equivalence between global contexts. 
4. `wf_exp_eq_gvar_ax_refl` should apply to any global variable. reflexivity should hold.
5. global context extensions should make sure the uniqueness of global names.

Ailrun:
1. I think this one and 2. are interrelated. If we have global substitutions, then we need to consider subtyping there. I am not sure whether this should interact with `wf_ctx_sub`, but if global substitutions interact with local substitutions, then I think its subtyping should also interact with `wf_ctx_sub`.
2. Why do you think we may need that?
3. Agreed
4. Agreed
5. Agreed

HuStmpHrrr:
I still don't think you need two global contexts in `wf_ctx_sub` as you never need to operate on two global contexts during type-checking. I don't seem able to come up with a case where global substitutions are used. if that's the case, global names are truly simple.

jiangsy:
1. it's actually implied, because the base case requires two identical Delta, I put it in this form mostly for symmetry, do you have a better notation if we use a single gctx?
2. I don't think we need gloabl substitutions either
3. same to 2
4. for non-axiom definitions, the reflexivity is implied by delta reduction, symmetry and transitivity, but I can add it explicitly
5. sure, I'll fix that

jiangsy:
fix 5 using stdpp's gset, but cause a notation conflicts: we have to rename all N to N' otherwise Coq confuses it with the inductive type and raise `Found an inductive type while a variable name was expected.`. 

I don't understand what exactly happened as N is actually defined in `Inductive Coq.Numbers.BinNums.N` but not stdpp. I presume stdpp modify the parsing of Coq, so terms can appear in more positions and leads to the ambiguity. 

I currently ad-hocly fix this by renaming all N to N' (and N' to N'' for sure). Do you have a better way to solve this?

Ailrun:
@jiangsy That `N` was not in the scope previously, but as the stdpp.gmap exports it, it is now in the scope.

jiangsy:
> 1. it's actually implied, because the base case requires two identical Delta, I put it in this form mostly for symmetry, do you have a better notation if we use a single gctx?

they're changed to `"Δ ⊢ Γ ⊆ Γ'"` and `"Δ ⊢ Γ ≈ Γ'"` now

jiangsy:
@Ailrun , @HuStmpHrrr could you review the current definition and I would like merge this first part

theories/Core/Syntactic/Syntax.v
```diff
@@ -24,6 +25,7 @@ Inductive obj : Set :=
           string -> obj ->                   (** x. Pf : M[x/x, x/y, refl A x/z] *)
           obj -> obj -> obj -> obj
 | var : string -> obj.
+(* TODO: do we need gvar for this? possibly yes *)
```
HuStmpHrrr, line 28: Before elaboration you can't know a variable is global or not, so I think no.

HuStmpHrrr:
Why is the CI failing?

jiangsy:
I didn't comment out all files, so some later properties are broken, but it has nothing to do with `Definition.v`

theories/Core/Syntactic/System/Definitions.v
```diff
@@ -22,464 +28,515 @@ Inductive ctx_lookup : nat -> typ -> ctx -> Prop :=
...
-     {{ Γ ⊢ M[σ∘τ] ≈ M[σ][τ] : A[σ∘τ] }} )
+  `( {{ Δ ;; Γ ⊢s τ : Γ' }} ->
+     {{ Δ ;; Γ' ⊢s σ : Γ'' }} ->
+     {{ Δ ;; Γ'' ⊢ M : A }} ->
+     {{ Δ ;; Γ ⊢ M[σ∘τ] ≈ M[σ][τ] : A[σ∘τ] }} )
+
+| wf_exp_eq_delta :
+  `( {{ ⊢ Δ ;; Γ }} ->
+     {{ `#x := [ M ] :: A ∈ Δ }} ->
+     (M = Some M') ->
```
Ailrun: Why do we have a separate equality here instead of inlining it?I don't think this causes an issue to do with UIP or so.

jiangsy: it was because of notation, i didn't register `Some _` and `None` in the exp domain, so they cannot be recognized in `{{ ... }}`, I can fix this

Ailrun: Ah, I see... it might be useful to have two more notations for the Definition case and Axiom case, but not sure.

theories/Core/Syntactic/Syntax.v
jiangsy: let's leave the comments here for now and revisit this when we deal with the algorithmic part

jiangsy:
I overload the `++` notation to swap the argument to be consistent with `,`

Ailrun:
@jiangsy To avoid confusion, I think it'd better to use a different name than `++`, especially because one can splice normal `++` using `^(...)` as well.

